PR #1, opened 2016-07-24
making it a shell script

KrishMunot:


bolinfest:
The absence of the file extension is intentional. This way, we can change it to Python, Ruby, etc. without renaming the script and breaking existing workflows.

PR #2, opened 2016-08-08
adding license url

KrishMunot:


bolinfest:
This is not the authoritative source of the license. https://github.com/facebookexperimental/eden/blob/master/LICENSE is.

KrishMunot:
Sorry, I rectified the mistake in an another PR - https://github.com/facebookexperimental/eden/blob/master/LICENSE

PR #19, opened 2020-06-06
Update .gitignore

singhaditya28:
For better security and stability

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,133 @@
 # Rust libraries and extensions
 Cargo.lock
 target/
+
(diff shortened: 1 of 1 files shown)


fanzeyi:
Sorry I'm afraid we could not take this. Our build system will not be generating these files under the project at all, and adding irrelevant Python-specific gitignore isn't really useful to this project. Thanks for contributing.

singhaditya28:
It has some OS based files which are neverin the project but created on the client side and sometimes they result in non - functioning of programs. Thus, please accept my PR, any addition/ deletion you want is welcomed

PR #28, opened 2020-07-16
pass --allow-system-packages to getdeps.py build of Mononoke

lukaspiatkowski:


lukaspiatkowski:
I've pinned the rust version to 1.44.0 for now

PR #45, opened 2020-08-26
configparser: fix non-fb dependencies

quark-zju:
Summary:
Fix referring to 'version' without proper codegen by making 'version' compile
without codegen. This fixes configparser test when version/src/lib.rs was not
generated.

Make unneeded deps without 'fb' feature optional.

This would hopefully fix the "EdenSCM Rust Libraries" GitHub workflow.

Reviewed By: DurhamG

Differential Revision: D23269864


quark-zju:
The configparser test is passing.

PR #71, opened 2021-02-03
change logging.warn to logging.warning

leszfb:
Summary: logging.warn() is deprecated since Python 3.3 in favor of logging.warning()

Reviewed By: ahornby

Differential Revision: D25870738


ahornby:
Can close this I think, can see same change has landed in master.

PR #77, opened 2021-03-16
rebaseconflict: initial step to support check-in conflict

quark-zju:
Summary:
Lots of things are missing compared to the regular rebase, ex:
- Rebasing multiple commits.
- Respecting the mutation markers.
- Rebasing merges.
- Automatic 3-way merges on conflicted files.
- Support from other commands like update, resolve.

But this demonstrates the basic concept.

Differential Revision: D26949659

PR #78, opened 2021-03-16
pyconflict: bindings for the conflict module

quark-zju:
Summary: Expose the Rust conflict module to Python.

Differential Revision: D26949660


quark-zju:
Duplicated with #77. I wasn't aware of exporting support stack.

PR #79, opened 2021-03-16
conflict: new crate for handling stacked conflicts

quark-zju:
Summary:
Main idea (use `(adds, removes)` two list to represent conflicts) comes from
https://github.com/martinvonz/jj. I still need to think about octopus merge,
but for regular 3-way merges or stacked 3-way merges I think I understand how
it works.

This diff defines the data model with some simple operations on them.

Note to Reviewers: Strongly suggest reading https://github.com/martinvonz/jj/issues/4#issuecomment-791179105 to understand how it works.

Differential Revision: D26949661


quark-zju:
Duplicated with #77. I wasn't aware of exporting support stack.

PR #81, opened 2021-03-22
[hg] update github actions to python 3

ahornby:


quark-zju:
You might want to change the `cargo test` commands to use Python 3 too. `cpython-ext` defaults Python 2 in its `Cargo.toml`:

    [features]
    default = ["python2"]
    python2 = ["cpython/python27-sys", "python27-sys"]
    python3 = ["cpython/python3-sys", "python3-sys"]

Something like `cargo test --no-default-features --features python3`.

Alternatively, you can change the `Cargo.toml`s so they use Python 3 by default. If you do that, `setup.py` and `setup3.py` also need to update to specify the right features.

ahornby:
rebased on top of commit that makes cpython-ext default to python3.

ahornby:
looks like getdeps doesn't currently pass through CPLUS_INCLUDE_PATH to boost's b2 build (or I haven't found out how to yet).   Considering adding an --env-allowlist to getdeps.py for env variable names we'll allow through to the underlying builder.

ahornby:
> looks like getdeps doesn't currently pass through CPLUS_INCLUDE_PATH to boost's b2 build (or I haven't found out how to yet). Considering adding an --env-allowlist to getdeps.py for env variable names we'll allow through to the underlying builder.
Update to python3.8 seemed to fix the boost-python build on mac

ahornby:
getdeps build fixes for python3 landed in master so closing this. there are still some githubs actions issues to resolve, but will revisit those in separated PR

PR #88, opened 2021-11-07
[draft] OSS builds

gabrielrussoc:
- Fixed branches from `master` to `main`.
- Removed FB only internal code
- Removed `fb` feature
- Some hacks to point to github crates instead of in-repo crates


eden/fs/config/EdenConfig.h
```diff
@@ -781,23 +781,10 @@ class EdenConfig : private ConfigSettingManager {
...
-
-  /**
-   * (Facebook Internal) Determines if EdenFS should use ServiceRouter.
-   */
-#ifdef EDEN_HAVE_SERVICEROUTER
-  ConfigSetting<bool> enableServiceRouter{
-      "facebook:enable-service-router",
-      fb_has_servicerouter(),
-      this};
-#else
```
fanzeyi, line 795: I don't think we define this for open source build? If it is defined we might want to fix that instead of removing this line.

eden/fs/config/EdenConfig.h
```diff
@@ -17,7 +17,7 @@
 #include <folly/portability/SysTypes.h>
 #include <folly/portability/Unistd.h>
 
-#include "common/rust/shed/hostcaps/hostcaps.h"
+//#include "common/rust/shed/hostcaps/hostcaps.h"
```
fanzeyi, line 20: Instead of removing it, maybe

```suggestion
#ifdef EDEN_HAVE_SERVICEROUTER
#include "common/rust/shed/hostcaps/hostcaps.h"
#endif
```

eden/scm/lib/backingstore/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 # This software may be used and distributed according to the terms of the
 # GNU General Public License version 2.
 
-rust_static_library(rust_backingstore CRATE backingstore FEATURES fb)
```
fanzeyi, line 6: Can you make this optional? I think we have a CMake variable to check if it is Facebook internal build or not.

fanzeyi (commented):
Thanks for the PR! I left a few comments. 

Curious: are you working on this on behalf of Databricks or in personal capacity?

gabrielrussoc:
@fanzeyi We (Databricks) are looking into ways to improve our monorepo performance and Eden looks promising. 

This PR was a step I took to validate if it actually works outside of FB and how supported it is.

gabrielrussoc:
@fanzeyi I interned at FB a few year ago too 😅

> The git support we have is very basic (if that's the route you are considering).

It's not very clear to me what kind of Git support there is. Let me open up an issue for better visibility and ask a few questions there

ahornby:
OSS builds are working again

PR #93, opened 2021-12-01
[EdenFs] update references to hardcoded rust shed location

dymart:
remove the references to common/rust that does not exist in this repo


ahornby:
I think the solution to these is to generate the cargo.toml with autocargo otherwise we'll end up with internal and external dependencies diverging.  I've got a diff underway internally to try to do this

dymart:
sounds great @ahornby! should I close this pr then?

ahornby:
Resolved by generating the Cargo.tomls with autocargo which generates the correct cross repo dependencies.   Current state is that Eden CLI and Mononoke test and build in GitHub actions.  Eden FS builds but the tests hang if enabled (maybe config related as same github checkout tests fine on internal host)

PR #95, opened 2021-12-01
[EdenFS] add conditional to backingstore cmakelists.txt

dymart:
Not sure if this is the best way to do this or what variable this should be conditional on. 
The main point is to remove the FEATURES fb from being called during an oss build


ahornby:
Resolved by using the IS_FB_BUILD flag from eden manifest, and by removing default "fb" feature from OSS Cargo.tomls

PR #96, opened 2021-12-01
[EdenFs] add conditional around hostcaps.h

dymart:


ahornby:
Resolved by removing default "fb" feature from OSS Cargo.toml

PR #118, opened 2022-04-12
Update README.md

jayaramcloud:
Added the text GNU General Public License v2.0 so that the users need not click LICENSE.


bolinfest:
We have rewritten the README.md for the Sapling launch, so this diff is out of date.

Further, while the bulk of the code in the repository is GPLv2, the `addons/` and `website/` folders are MIT licensed, so this is not completely accurate.

PR #152, opened 2022-11-16
Adjust `git reset --hard` in Git cheat sheet

### LINKED ISSUE #150: Equivalence between `git reset` and `sl hide` is wrong
In the [Git cheat sheet](https://sapling-scm.com/docs/introduction/git-cheat-sheet), the following are equated:

- `git reset --hard COMMIT`
- `sl hide COMMIT`

This is wrong. The former resets *to* `COMMIT`, essentially deleting commits up to but not including `COMMIT`, while `sl hide` will also delete the target commit.

arxanas:
Closes #150

### DIFF
--- a/website/docs/introduction/git-cheat-sheet.md
+++ b/website/docs/introduction/git-cheat-sheet.md
@@ -64,7 +64,7 @@ See `sl help revset` for more ways of referencing commits.
 |Copy file |cp OLD NEW |sl cp OLD NEW |
 |Add/remove all files |git add -A . |sl addremove |
 |Undo changes |git checkout -- FILE |sl revert FILE |
-|Undo all changes |git reset --hard |sl revert —all |
+|Undo all changes |git reset --hard |sl revert --all |
 |Delete untracked files |git clean -f |sl clean |
 |Output file content |git cat-file -p COMMIT:FILE |sl cat -r COMMIT FILE |
 |Show blame |git blame FILE |sl blame FILE |
@@ -94,7 +94,7 @@ See `sl help revset` for more ways of referencing commits.
 |Undo commit |git reset --soft HEAD^ |sl uncommit |
 |Undo partial commit |git reset --soft HEAD^ FILE |sl uncommit FILE |
 |Undo amend |git reset HEAD@{1} |sl unamend |
-|Undo rebase/etc |git reset —hard HEAD@{1} |sl undo |
+|Undo rebase/etc |git reset --hard HEAD@{1} |sl undo |
 |Revert already landed commit |git revert COMMIT |sl backout COMMIT |
 |View recent commits |git reflog |sl journal |
 |Recover commit |git reset COMMIT |sl unhide COMMIT |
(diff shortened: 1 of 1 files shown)


arxanas:
By the way, I couldn't figure out how to get `sl pr` to work, so I submitted this manually 😦.

```
 ~/Workspace/sapling $  sl                                                                                             ◉
@  42e8d98a8  19 minutes ago  me  git-cheat-sheet* remote/git-cheat-sheet
╷  Adjust Git cheet sheet `git reset --hard` equivalence
╷
o  fe5ab6dcf  Today at 14:24  remote/main
│
~
 ~/Workspace/sapling $  sl pr                                                                                          ◉
no commits to submit
```

website/docs/introduction/git-cheat-sheet.md
```diff
@@ -94,7 +94,7 @@ See `sl help revset` for more ways of referencing commits.
 |Undo commit |git reset --soft HEAD^ |sl uncommit |
 |Undo partial commit |git reset --soft HEAD^ FILE |sl uncommit FILE |
 |Undo amend |git reset HEAD@{1} |sl unamend |
-|Undo rebase/etc |git reset —hard HEAD@{1} |sl undo |
+|Undo rebase/etc |git reset -—hard HEAD@{1} |sl undo |
```
bolinfest, line 97: I'm looking at this on ReviewStack (https://reviewstack.dev/facebook/sapling/pull/152) and it appears to be non-ASCII?

website/docs/introduction/git-cheat-sheet.md
```diff
@@ -76,7 +76,7 @@ See `sl help revset` for more ways of referencing commits.
 |Commit changes |git commit -a |sl commit |
 |Modify commit |git commit -a --amend |sl amend |
 |Move to commit |git checkout COMMIT |sl goto COMMIT |
-|Remove a commit |git reset -hard COMMIT |sl hide COMMIT |
+|Remove current commit |git reset --hard HEAD^ |sl hide . |
```
bolinfest, line 79: I think it's important to include the more general "Remove a commit" use case, though.

website/docs/introduction/git-cheat-sheet.md
arxanas: @bolinfest I see, there were previously em-dashes in the file, and I just added a single hyphen, thinking that they were regular dashes. I've updated the commit to instead replace all em-dashes.

arxanas: I checked, and there are only a handful of other en- and em-dashes in the repository, but only in places where it doesn't matter for copy/paste or execution reasons.

website/docs/introduction/git-cheat-sheet.md
arxanas: @bolinfest What do you propose for the more general "remove a commit" case in Git? There isn't really an easy way to do that. I suppose you could try to use `git rebase -i` from all branches for which the commit is reachable and manually remove it, or, if the commit is the tip of some branch(es), you can use `git branch -f`/`git reset --hard` (depending on whether or not the branch is checked out 🙄). As the example is previously written, it wasn't removing a commit in the general case, anyways.

The Sapling approach to removing a commit isn't fully general, either: in my experimentation, it will also hide all descendant commits, so you can't remove just one commit in the middle of a stack.

quark-zju: I think the current way is fine. Removing an arbitrary commit might require knowledge about the graph topology, which seems hard to explain in a small table cell.

PR #170, opened 2022-11-16
Fix missing extern for identity crate

yonas:


yonas:
Posted this too early. I'm now getting a `cant' find crate` error.

bolinfest:
I'm not sure what issue you're hitting. Possibilities:

- Do you need to `make clean` and then `make oss` again?
- Are you on an M1? For our prebuilt binaries, we cross-compile from Intel Mac, so building on M1 directly isn't tested quite as well.

quark-zju:
Are you using an old rustc? `extern crate` became unnecessary since Rust edition 2018: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-extern-crate and we're targeting Rust edition 2021 now. I'm closing this since upgrading rustc would solve this issue.

PR #199, opened 2022-11-18
[sl] ci: add circleci integration for macOS build

sggutier:
Stack created with [Sapling]
* __->__ #199

[sl] ci: add circleci integration for macOS build


bolinfest:
Can we incorporate this into the gen workflows script since the first two jobs need to stay in sync?

PR #219, opened 2022-11-19
Show pull request unit-test status in smartlog

shish:
I want to be able to see if my test suite has passed all checks from the comfort of my smartlog

I don't actually know if this works yet because I haven't managed to get sapling building _at all_, let alone with my changes -- I'm submitting preemptively to check if the feature would be welcome, and if so, am I going in the right direction?

(Also while I take however many hours it takes to set up a working local build, if somebody else were to take this code as a starting point and submit a working PR, that'd be cool too :) )


shish:
Closing this as I got it working in another branch

PR #222, opened 2022-11-20
hardcode github.com for STACK_HEADER

discentem:
It does not seem likely that `ezyang/ghstack` will exist at `https://{github_url}/ezyang/ghstack)`. I assume that the hyperlink is meant to give credit/reference to ezyang's project, so this fixes the hyperlink so that it always works even if we've clone a repo from a GHE instance.


bolinfest:
Good catch!

Issue #258, opened 2022-11-24
`sl pr` crashes at first use


mimoo:
following the tutorial here: https://sapling-scm.com/docs/introduction/getting-started

my guess is that my system's python doesn't work with sl?

```
$ sl pr
** Sapling SCM (version 0.1.20221118-210929-cfbb68aa) has crashed:
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/__init__.py", line 106, in run
    dispatch.run(args, fin, fout, ferr, config)
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/dispatch.py", line 161, in run
    status = (dispatch(req) or 0) & 255
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/dispatch.py", line 503, in dispatch
    ret = _runcatch(req)
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/dispatch.py", line 712, in _runcatch
    return _callcatch(ui, req, _runcatchfunc)
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/dispatch.py", line 721, in _callcatch
    return scmutil.callcatch(ui, req, func)
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/scmutil.py", line 147, in callcatch
    return func()
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/dispatch.py", line 683, in _runcatchfunc
    return _dispatch(req)
  File "/usr/local/Cellar/python@3.8/3.8.15/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/edenscm/dispatch.py", line 1218, in _dispatch
    ret = runcommand(
[...]

PR #259, opened 2022-11-24
fixing default branch name for brand new github repos

mimoo:
attempts to fix https://github.com/facebook/sapling/issues/260 and https://github.com/facebook/sapling/issues/258

but I can't test this locally due to https://github.com/facebook/sapling/issues/261

### DIFF
--- a/eden/scm/edenscm/ext/github/gh_submit.py
+++ b/eden/scm/edenscm/ext/github/gh_submit.py
@@ -149,12 +149,13 @@ query ($owner: String!, $name: String!, $number: Int!) {
 
 
 def _parse_repository_from_dict(repo_obj, hostname: str, upstream=None) -> Repository:
+    default_branch = "main" if "defaultBranchRef" not in repo_obj else repo_obj["defaultBranchRef"]["name"]
     return Repository(
         id=repo_obj["id"],
         hostname=hostname,
         owner=repo_obj["owner"]["login"],
         name=repo_obj["name"],
-        default_branch=repo_obj["defaultBranchRef"]["name"],
+        default_branch=default_branch,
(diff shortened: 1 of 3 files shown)


bolinfest:
Can you please use `sl pr` or some other tooling to make this three separate pull requests?

eden/scm/edenscm/ext/github/gh_submit.py
```diff
@@ -149,12 +149,13 @@ async def get_pull_request_details(
 
 
 def _parse_repository_from_dict(repo_obj, hostname: str, upstream=None) -> Repository:
+    default_branch = "main" if "defaultBranchRef" not in repo_obj else repo_obj["defaultBranchRef"]["name"]
```
bolinfest, line 152: I think that we should be doing something like the following instead:

https://github.com/facebook/sapling/blob/6b7c5a62d1a60c4eb28e4f92e1f02c13f8c8f567/eden/scm/ghstack/github_utils.py#L91-L100

Though now that I add a pointer to that code, I notice that it needs to be updated to support GitHub Enterprise.

rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.65.0"
```
bolinfest, line 2: I haven't used this before: can this be a semver range or is "1.65.0 or later" implied?

bolinfest (changes requested):

bolinfest:
FYI, I'm going to try something like https://github.com/facebook/sapling/pull/263 to handle the case when no default branch exists.

bolinfest:
I addressed the middle commit via https://github.com/facebook/sapling/pull/263.

PR #264, opened 2022-11-24
[docs] add sl-shell-example syntax-highlighting language

shish:
Stack created with [Sapling]
* #270
* #267
* #266
* __->__ #264

[docs] add sl-shell-example syntax-highlighting language

Summary: Right now the example sl sessions have no hightlighting, or sometimes they pretend to be bash, which kind-of-half-works but also half doesn't work and is ugly (See the "Before" example - there is random highlighting for some numbers in the middle of command output, the ".." in "..." is randomly highlighted as a directory, etc)

This diff adds a new custom Prism language called "sl-shell-example" which does syntax highlighting specifically tailored for our example sessions with these main parts:

```
# comment
$ shell command
~/my/subdir $ shell command in a specific subdirectory
output
```

Screenshot 2022-11-24 at 16 15 08

(For the record, Prism already has a language called "shell-session", but that language doesn't support comments (which is something we make heavy use of), and also it tries to be clever by highlighting random words in the middle of shell commands (which is just distracting in the context of sapling docs))

Test Plan:

Before:
Screenshot 2022-11-24 at 15 54 27

After:
Screenshot 2022-11-24 at 15 54 39

### DIFF
--- a/website/docs/internals/drawdag.md
+++ b/website/docs/internals/drawdag.md
@@ -10,7 +10,7 @@ When creating tests, we often need to create a repo with a particular layout.
 For example, to create a linear graph with three commits, we could use the
 following sequence of commands:
 
-```
+```sl-shell-example
 $ sl commit -m A
 $ sl commit -m B
 $ sl commit -m C
@@ -18,7 +18,7 @@ $ sl commit -m C
 
 If the graph is nonlinear, extra commands such as merge and goto are needed:
 
-```
+```sl-shell-example
 $ sl commit -m A
 $ sl commit -m B
 $ sl goto -q '.^'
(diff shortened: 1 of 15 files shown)


shish:
If somebody is some kind of regex-genius it might even be possible to automatically highlight smartlog examples, but I'll leave that as an exercise for the reader (or my future self :) )

mzr:
Thanks for the PR! LGTM, but one nit from me. Can we change `sl-session` to something like `sl-shell-example`? IMO it better reflects what it really is, and it is also similar to original `shell-session` in Prism.

shish:
> Can we change `sl-session` to something like `sl-shell-example`?

Done :)

Issue #228, opened 2022-11-20
`sl init` does not work on existing directories.


rbehrends:
Unlike `git init` or `hg init`, `sl init --git DIR`  results in an error if `DIR` already exists. This is even more puzzling as the directory argument is optional and defaults to the current directory, but you cannot actually omit it, because the current directory always exists.

A relevant use case is putting existing code under version control, i.e. doing `sl init --git` within an existing project directory.

A workaround is to do `sl init --git some_new_tmpdir` and then manually `mv` the `.sl` directory to the current directory and remove the temporary directory.

Ivan-Sanchez-Diaz:
sorry for the duplicate pull request! first contribution attempt :sweat_smile:

bolinfest:
This is fixed in https://github.com/facebook/sapling/commit/a4dbd37fa70fa182df8b1a824f2231b37e417830 and will therefore go out in the next release.

PR #277, opened 2022-11-27
fix ``sl init`` in existing directories

Ivan-Sanchez-Diaz:
as mentioned in #228 ``sl init --git DIR`` results in an error if ``DIR`` already exists, this fix creates a temporary directory if the directory exists and then moves the ``.sl`` directory into the target directory, if the target directory exists and already contains a ``.sl`` directory an error is thrown saying ``destination DIR is already a repository``


eden/scm/edenscm/git.py
```diff
@@ -111,11 +112,18 @@ def clone(ui, url, destpath=None, update=True, pullnames=None):
         destpath = os.path.realpath(basename)
 
     destpath = ui.expandpath(destpath)
-    if os.path.lexists(destpath):
-        raise error.Abort(_("destination '%s' already exists") % destpath)
+    if os.path.lexists(destpath) and os.listdir(destpath):
+        raise error.Abort(_("destination '%s' is not empty") % destpath)
+    
+    exists = os.path.lexists(destpath)
```
bolinfest, line 118: Let's only compute this once.

eden/scm/edenscm/git.py
```diff
@@ -111,11 +112,18 @@ def clone(ui, url, destpath=None, update=True, pullnames=None):
...
+    
+    exists = os.path.lexists(destpath) 
 
     try:
-        repo = createrepo(ui, url, destpath)
+        if exists:
+            temppath = tempfile.mkdtemp()
+            repo = createrepo(ui, url, temppath )
+        else: 
+            repo = createrepo(ui, url, destpath)
```
bolinfest, line 125: Does `createrepo()` not work if `destpath` is an empty folder?

eden/scm/edenscm/git.py
```diff
@@ -128,6 +136,10 @@ def clone(ui, url, destpath=None, update=True, pullnames=None):
         repo = None
         shutil.rmtree(destpath, ignore_errors=True)
         raise
+    
+    if exists: 
+        shutil.move(temppath +"/.sl",destpath)
+        shutil.rmtree(temppath, ignore_errors=True)
```
bolinfest, line 142: Should `temppath` by empty if `shutil.move()` succeeded?

bolinfest (changes requested):
I want to confirm that we still need special logic if the destination folder was empty to start with.

eden/scm/edenscm/git.py
Ivan-Sanchez-Diaz: yes, the directory is empty after ``shutil.move()`` and therefore ``os.rmdir()`` could be used but since ``createrepo()`` works on empty folders this is no longer necessary

PR #324, opened 2022-12-13
Add UI Icon on the editor bar for open-isl command

G3orge26:
Hi Everyone. This is not really any ground breaking change but I added an icon using the .svg logo of sapling on the title bar of the editor. I think this is a nice alternative to hitting F2 and typing the command. Ideally it would be better if it was on the side bar as an Activitybar ( why not both ? ) but it appears that is not something trivial ( still havent found a way to make it work)

Screenshot 2022-12-13 184053

Tested by running yarn build-extension and copying the files outputed in dist/ as well as the project.json on the extensions folder of the vscode installation.

### DIFF
--- a/addons/vscode/package.json
+++ b/addons/vscode/package.json
@@ -48,7 +48,11 @@
       {
         "command": "sapling.open-isl",
         "title": "%sapling.open-isl-command%",
-        "category": "%sapling.command-category%"
+        "category": "%sapling.command-category%",
+        "icon": {
+          "light": "resources/Sapling_favicon-light-green.svg",
+          "dark": "resources/Sapling_favicon-light-green.svg"
+        }
       },
       {
         "command": "sapling.open-file-diff-uncommitted",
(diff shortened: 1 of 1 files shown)


bolinfest:
@G3orge26 Thanks! In general, this looks good, but I want to play with it first before merging.

bolinfest:
@G3orge26 upon further investigation, it is not possible to scope the `when` for `editor/title` as `"scmProvider == sapling"`. Unfortunately, this means that this icon will display even when the file is not part of a Sapling repo.

I don't think this will be a great user experience because it will create a lot of noise for folks who are working in a Git repo while using VS Code.

Let's take a step back: what is the goal here? If it is to provide a way to launch ISL from a button rather than the command palette, I think we should explore alternatives.

bolinfest:
Closing in favor of https://github.com/facebook/sapling/commit/538c6fba11ddfdae9de93bf77cffa688b13458c0

PR #327, opened 2022-12-14
foo

bolinfest:
foo


bolinfest:
I could write some comments here.

foo
```diff
@@ -0,0 +1 @@
+foo
```
bolinfest, line 1: line comment!

PR #340, opened 2022-12-20
surface output in addition to exit code if ghstack shell command fails

discentem:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/340).
* __->__ #340

surface output in addition to exit code if ghstack shell command fails

### DIFF
--- a/eden/scm/ghstack/shell.py
+++ b/eden/scm/ghstack/shell.py
@@ -218,8 +218,8 @@ class Shell(object):
             return returncode == 0
         if returncode != 0:
             raise RuntimeError(
-                "{} failed with exit code {}"
-                .format(' '.join(args), returncode)
+                "{} failed with exit code {}. Output was '{}'"
+                .format(' '.join(args), returncode, out)
             )
 
         if stdout == subprocess.PIPE:


bolinfest:
@discentem I'm having some trouble importing this into our tooling. Could you try rebasing on `main` and submitting this PR again? I would expect to see only one commit.

discentem:
Curious. But no problem, I sh  I'll try to look into it later today.

bolinfest:
@discentem checking with folks on our side, we believe the issue is that https://github.com/facebook/sapling/pull/340/commits/d0fc2c21407ec112e543b7f0bf74a2fd0a157592 does not have a "valid author," i.e., no GitHub info attached, which is not the case with the other commit that is also authored by "you" in the stack, but a different person from the Author line perspective?

discentem:
I think what happened is that my fork was out of date. So I synced the fork in Github UI, pulled remote/main again, and rebased on top of it. Looks like there is now only one commit in this PR 😅

zpao:
I think the problem is actually with the commit itself. You likely have a problem with your `.gitconfig`, but you can at least fix it by amending the commit here. Your author line is `"BK EMAIL" <>` (where `EMAIL` looks to be your real email, so not writing it here) which looks wrong. You probably want `"BK <EMAIL>"`. So `git commit --amend --author="BK <EMAIL>"` should do it.

bolinfest:
@discentem I believe that @zpao is right in that our import tool is still not working due the autho rline, even though the PR contains only a single commit now.

Assuming you are in a Sapling repo, `sl metaedit --user "NAME <EMAIL>"` *should* fix it, though I noticed this appears to update the `author` field for a commit, but not the `committer` field. Not sure whether this is intentional or an oversight (/cc @quark-zju).

Note you can see the raw commit by running the following from the root of your Sapling repo:

```
git --git-dir .sl/store/git cat-file -p $(sl whereami)
```

discentem:
@bolinfest I did as you suggested `sl metaedit --user "BK <MY_ACTUAL_EMAIL>"` (which seems equivalent to `sl amend -u "BK <MY_ACTUAL_EMAIL>`) but this still leaves "a separate user" mentioned in the authoring bit of the commit

[discentem](https://github.com/facebook/sapling/commits?author=discentem) authored and BK MY_OTHER_EMAIL committed 1 hour ago

🤔 

Sorry for so much trouble for a trivial commit 🙃 😫

bolinfest:
Yep, I see...well, until we change `metaedit`, how about this:

Do `sl histedit` and change the command for the line to `edit` and save.
That should put things in an unstaged state.

Then do:

```
sl commit --config ui.username="USER <EMAIL>" -m 'previous commit message'
```

so that should be the same contents as before, but with the author/email as the `ui.username` on the command line.

Then do:

```
sl link https://github.com/facebook/sapling/pull/340
```

to associate it with this PR and then `sl ghstack` or `sl pr submit` depending on what you did originally (I think the latter)?

quark-zju:
> though I noticed this appears to update the author field for a commit, but not the committer field

It is not intentional. We should probably change the behavior and/or make it configurable.

discentem:
Thanks @zpao, @bolinfest, and @quark-zju for all the assistance here, much appreciated 😄

discentem:
@quark-zju If you can point me at the source files where this stuff happens I'd love to take a stab at fixing the committer vs author thing

discentem:
> @quark-zju If you can point me at the source files where this stuff happens I'd love to take a stab at fixing the committer vs author thing

Nevermind, think I found it: https://github.com/facebook/sapling/blob/e8f57d7902c800bec7ded3c567d9e370460df24e/eden/scm/edenscm/ext/amend/metaedit.py#L103

quark-zju:
> @quark-zju If you can point me at the source files where this stuff happens I'd love to take a stab at fixing the committer vs author thing

I can fix this. We probably want to fix it in the right place so it applies to all commit rewrite operations like `rebase` etc, not just for `metaedit`.

discentem:
> > @quark-zju If you can point me at the source files where this stuff happens I'd love to take a stab at fixing the committer vs author thing
> 
> 
> 
> I can fix this. We probably want to fix it in the right place so it applies to all commit rewrite operations like `rebase` etc, not just for `metaedit`.

That definitely makes a lot of sense. Out of curiosity, is there a central place where the rewrite logic is? I searched around a lot and could not find where this might live.

bolinfest:
@discentem FYI @quark-zju has a fix with a test in flght!

PR #344, opened 2022-12-21
surface output in addition to exit code if ghstack shell command fails

discentem:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/344).
* __->__ #344

surface output in addition to exit code if ghstack shell command fails


discentem:
accidental dupe of #340

PR #354, opened 2022-12-24
add up & down as aliases for next & prev

discentem:
add up & down as aliases for next & prev

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/354).
* __->__ #354

### DIFF
--- a/eden/scm/edenscm/ext/amend/movement.py
+++ b/eden/scm/edenscm/ext/amend/movement.py
@@ -28,7 +28,7 @@ moveopts = [
 
 
 @command(
-    "previous|prev",
+    "previous|prev|last|down",
     [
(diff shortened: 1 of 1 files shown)


bolinfest:
I'm not sure we want to take this because I'm hesitant of having multiple names for the same action. Generally, this makes things harder to document and can lead to confusion because someone has to confirm if it is the same thing they already do.

You can always add your own alias for this:

```
sl config --user alias.up prev
```

discentem:
Oh! a personal alias works just fine for me. I didn't realize all the different things I could set in with `config` 😅

Issue #363, opened 2022-12-27
`sl pr list` exits before returning data


francois:
Running `sl pr list` returns to the command line, then a few moments later (< 1 second), the PR list is printed:

```
╰─○ sl pr list
╰─○
Showing 2 of 2 open pull requests in Inspectify/marketplace

#732  Some title  pr732                         about 3 days ago
#622  Some title  feature/remove_success_again  about 28 days ago
```

```
╰─○ sl --version
Sapling 0.2.20221222-152408
╰─○ brew --version
Homebrew 3.6.16
Homebrew/homebrew-core (git revision 4c7d9dfaa44; last commit 2022-12-27)
Homebrew/homebrew-cask (git revision 82cb73dc4c; last commit 2022-12-27)
╰─○ gh --version
gh version 2.21.1 (2022-12-23)
https://github.com/cli/cli/releases/tag/v2.21.1
```

muirdm:
Thanks for the report. It looks related to the pre-fork server used to speed up commands, combined with the exec of the "gh" command to list prs. An upcoming change should resolve this.

bolinfest:
@quark-zju I am having a lot of trouble finding the code that maps requests from the client to the server for `ui.system()`. I see this comment:

https://github.com/facebook/sapling/blob/d0936284bd7208707ccf57a83a8faf7b26e587e5/eden/scm/edenscm/chgserver.py#L15-L16

and this code:

https://github.com/facebook/sapling/blob/d0936284bd7208707ccf57a83a8faf7b26e587e5/eden/scm/edenscm/chgserver.py#L130

But I don't see where the server maps `b"s"` to `system()` or `_runsystem()` or whatever it is supposed to be.

Trying to escape arguments and concatenate them with spaces to create an equivalent `cmd` as a single string is quite a pain because `shlex()` is not designed to support Windows. I want to add an RPC that takes the args as an array and invokes `subprocess.call()` with `shell=False`.

quark-zju:
> But I don't see where the server maps `b"s"` to `system()` or `_runsystem()` or whatever it is supposed to be.

Is https://github.com/facebook/sapling/blob/d0936284bd7208707ccf57a83a8faf7b26e587e5/eden/scm/contrib/chg/hgclient.c#L230 what you look for?
 
[...]

PR #377, opened 2022-12-31
use ui.system() instead of os.execv() to run `gh list`

bolinfest:
use ui.system() instead of os.execv() to run `gh list`

Summary:

As described in https://github.com/facebook/sapling/issues/363,
the existing logic did not work correctly in the presence of chg.

Test Plan:
Build Sapling from source and verify that `sl pr list` works with
various arguments.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/377).
* __->__ #377


bolinfest:
Oh, I should probably check the exit code, eh?

PR #440, opened 2023-01-08
(build): Use Sapling HEAD for sl version by default in `make oss`

vegerot:
(build): Use Sapling HEAD for sl version by default in `make oss`

Summary: Currently, following the recommended documentation for building
Sapling will always set the sapling version to the upstream Mercurial version
(which has been 4.4.2 for years)

```sh
❯ sl --version
Sapling 4.4.2
```

This isn't very useful when debugging Sapling, and is inconsistenct with the
distributed versions of sl.  This commit aligns the sl version with the sl
version in the distributed versions of sl

Test Plan: manual

```sh
❯ (cd eden/scm && make oss && ./sl --version)
Sapling 0.2.20230107-190820-h75ace1ad
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/440).
* #522
* __->__ #440

### DIFF
--- a/eden/scm/setup.py
+++ b/eden/scm/setup.py
@@ -311,7 +311,7 @@ def filterhgerr(err):
     return b"\n".join(b"  " + e for e in err)
 
 
-def findhg():
+def findhg(vcname: str):
     """Try to figure out how we should invoke hg for examining the local
     repository contents.
 
@@ -328,7 +328,7 @@ def findhg():
     # and disable localization for the same reasons.
     hgenv["HGPLAIN"] = "1"
     hgenv["LANGUAGE"] = "C"
-    hgcmd = ["hg"]
+    hgcmd = [vcname]
     # Run a simple "hg log" command just to see if using hg from the user's
     # path works and can successfully interact with this repository.
     check_cmd = ["log", "-r.", "-Ttest"]
@@ -341,7 +341,7 @@ def findhg():
 
     # Fall back to trying the local hg installation.
     hgenv = localhgenv()
-    hgcmd = [sys.executable, "hg"]
+    hgcmd = [sys.executable, vcname]
     try:
         retcode, out, err = runcmd(hgcmd + check_cmd, hgenv)
     except EnvironmentError:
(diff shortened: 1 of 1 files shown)


sggutier (changes requested):
Thanks for the interest on changing the versioning schema for debugging purposes.

What if instead of the changes you made here you [modify the `pickversion` function](https://github.com/facebook/sapling/blob/97b87376653b459cd312d95a0ceb6ef62f416191/eden/scm/setup.py#L388-401) on `setup.py`? 

You could change it so that it uses `sl` instead of `hg` if it is available on the OSS build (or maybe even git).

eden/scm/Makefile
```diff
@@ -40,6 +40,8 @@ OSXVERSIONFLAGS ?= $(shell echo $$OSXVERSIONFLAGS)
 HGNAME ?= hg
 SL_NAME = sl
 
+export SAPLING_VERSION ?= $(shell ../../ci/tag-name.sh)
```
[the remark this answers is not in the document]
vegerot: updated!  What do you think?

sggutier:
Looks better now (and I would really like to approve it), but there is still one troublesome issue

eden/scm/setup.py
```diff
@@ -372,9 +372,7 @@ def localhgenv():
         env["SystemRoot"] = os.environ["SystemRoot"]
     return env
 
-
```
sggutier, line 375: Prescribing a VCS without first detecting the repo type (hg, sl, git) still bugs me out. Could something be done for detecting the repo type?

Also, users having both hg and sl installed might be more common than one might think.

quark-zju: I think this is an improvement. Previously `setup.py` hardcodes `"hg"` which obviously won't work in all cases. Whether we detect `.hg` or `.sl` or just try out both of them and pick one that works seems to be just an implementation detail to achieve the same goal, and I don't think that should block this PR.

Issue #447, opened 2023-01-11
Cannot run tests locally


vegerot:
When trying to run or add tests, I run into a variety of errors.  Many of which reference fb tools, meaning that the test suite likely is not set up to be run outside of Meta.

I'll post more details about the tests and the errors later 🙂

strager:
#458 should help some.

vegerot:
tl;dr:
``sh
$ cd tests
$ python run-tests.py test-<name>.t
```

However, most tests do not pass.  This issue includes trying to make the full test suite pass on the OSS build

PR #448, opened 2023-01-11
feat(scm): Add revset aliases for `next` and `prev`

### LINKED ISSUE #444: [proposal] sl goto top/bottom/prev/next ?
I have found myself trying to run `sl goto --bottom` and `sl goto --top` a few times. I know `--bottom` and `--top` are supposed to be used with `next` and `prev`, but my fingers do the typing, not my brain.

I would argue that for consistency, having a single top-level gateway into moving the current tip would be preferable:

```
sl goto next # sl next
sl goto prev # sl prev
sl goto top # sl next --top
sl goto bottom # sl prev --bottom
sl goto $SHA1 # same as currently
sl goto $BOOKMARK # same as currently
```

I also know that currently, nothing prevents a bookmark from being named `top`, `bottom`, `next` or `prev`. Those would have to become reserved words in bookmarks.

What do you think?

vegerot:
feat(scm): Add revset aliases for `next` and `prev`

Summary: In the same way you can `sl go top`, it seems logical to also have
`sl go next`. However, instead it is `sl next`.  While I normally remember to
do `sl prev`, I sometimes accidentally do `sl go prev` out of habbit and get an
error.  Instead, let's simply add two revset aliases for `next` and `prev` to
make sl more intuitive.

Differences from `sl next/prev`:
- Cannot specify number of commits: `sl prev 2` works but `sl goto prev 2`
doesn't. (but `sl go prev~1` would)
- Handles merge commits differently.  No interactive selection and no `sl next --newest`

Closes #444

Test Plan: Automated tested blocked on #447

Manually tested by running `sl go next` and `sl go prev` and verifying I ended
up in the right spot.

Question for reviewer: I don't know if this is relevant to this patch or not
but while testing this I would sometimes have this error

```py
./sl
Traceback (most recent call last):
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/__init__.py", line 86, in run
    dispatch.runchgserver(args[2:])
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/dispatch.py", line 368, in runchgserver
    _preimportmodules()
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/dispatch.py", line 348, in _preimportmodules
    extensions.preimport(extname)
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/extensions.py", line 208, in preimport
    mod = getattr(__import__("edenscm.ext.%s" % name).ext, name)
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/ext/copytrace.py", line 169, in <module>
    ) -> Tuple[Optional[dbm._Database], Optional[Tuple[Type[dbm._error], Type[OSError]]]]:
AttributeError: module 'dbm' has no attribute '_Database'
```

and other times I would get this warning

```python
CHGDISABLE=1 ./sl go prev
Traceback (most recent call last):
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/extensions.py", line 399, in _runextsetup
    extsetup(ui)
  File "/Users/m0c0j7y/workspace/github.com/facebook/sapling/eden/scm/edenscm/ext/copytrace.py", line 113, in extsetup
    extensions.wrapfunction(copiesmod, "mergecopies", _mergecopies)
NameError: name '_mergecopies' is not defined
warning: failed to set up extension copytrace: name '_mergecopies' is not defined
1 files updated, 0 files merged, 0 files removed, 0 files unresolved
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/448).
* __->__ #448


discentem:
@vegerot is this the same as https://github.com/facebook/sapling/pull/354? If so, the maintainers unfortunately didn't want to go for it. 

And does the suggestion of aliases apply here? https://github.com/facebook/sapling/pull/354#issuecomment-1367576823

vegerot:
> @vegerot is this the same as #354? If so, the maintainers unfortunately didn't want to go for it.
> 
> And does the suggestion of aliases apply here? [#354 (comment)](https://github.com/facebook/sapling/pull/354#issuecomment-1367576823)

I don't think this is the same kind of thing.  Making `sl go next` behave like `sl next` imo is a consistency thing.

> I'm hesitant of having multiple names for the same action. Generally, this makes things harder to document and can lead to confusion because someone has to confirm if it is the same thing they already do.

@bolinfest makes a really good point here I'll need to think on.  Certainly, in my experience this is true.  Both as a user I have to look up if `fold` is the same as `squash`, and as a developer I have to maintain both of them.  Additionally, this takes up an extra "keyword" we couldn't use for something else later on.

I think it's a tradeoff because having multiple names for the same thing (which this commit is _not_) can also make the program more intuitive.

Again, I don't think this commit is adding "multiple names for the same thing".  Instead, it is adding a revset alias for a common revision.

We already have a rev alias for `bottom` so I can do things like

```
sl log --rev bottom::.
```

Since we felt that `bottom` was convenient enough, then why not let me do 

```
sl log --rev bottom::next
```
?

discentem:
> Again, I don't think this commit is adding "multiple names for the same thing". Instead, it is adding a revset alias for a common revision.

Ack thanks for explaining, yep it doesn't seem like the same thing.

strager:
I would like this feature! `sl histedit next` would be a nice alternative to `sl histedit .~-1`, for example. (The latter has too much symbol soup.)

@vegerot Does this PR handle ambiguities (multiple parents (merge commits); multiple children) the same way that `sl prev` and `sl next` do?

vegerot:
> @vegerot Does this PR handle ambiguities (multiple parents (merge commits); multiple children) the same way that `sl prev` and `sl next` do?

`sl next` is implemented entirely differently in `movement.py`.  When beginning this patch I took a gander through `_moverelative` but decided to pursue the simpler implementation of `.~1`.

If by "ambiguities" you mean interactively selecting merged parents or the `--newest` flag then no.  I forgot to write about this in my commit.  But FYI `prev` behaves exactly the same as `.~1`.  You can test how this'll behave by using `.~1`.

> I would like this feature! `sl histedit next` would be a nice alternative to `sl histedit .~-1`, for example. (The latter has too much symbol soup.)

If you like this, you can try this out now (without waiting for this to be merged) by adding to your config (`sl config --user`)

```toml
[revsetalias]
next=.~-1
prev=.~1
```

quark-zju:
Note I updated the logic so it reuses the `movement.py` logic to handle ambiguities.

vegerot:
thank you!! 🙏

PR #451, opened 2023-01-13
Fedora rpm changes

kiilerix:
Various minor changes to the sapling repo while hacking on Fedora packaging.

These changes works with https://gist.github.com/kiilerix/b32cb9688d2d6e9e752e91e2ca4c30e6 .

Still a long way to go to be able to package offline and reproducible.

### DIFF
--- a/eden/scm/setup.py
+++ b/eden/scm/setup.py
@@ -513,7 +513,7 @@ void print_buildinfo() {
 """
 
     path = pjoin(builddir, "buildinfo.c")
-    write_if_changed(path, buildinfosrc)
+    write_if_changed(path, buildinfosrc.encode())
     return path
 
 
--- a/eden/scm/setup.py
+++ b/eden/scm/setup.py
@@ -1305,7 +1305,8 @@ class hginstalllib(install_lib):
         for src, dst in [("edenscmdeps3.zip", "edenscmdeps3.zip")]:
             srcpath = pjoin(builddir, src)
             dstpath = pjoin(self.install_dir, dst)
-            file_util.copy_file(srcpath, dstpath)
+            if os.path.exists(srcpath):
+                file_util.copy_file(srcpath, dstpath)
 
 
 class hginstallscripts(install_scripts):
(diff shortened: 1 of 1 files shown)


bolinfest:
I haven't tested this yet, but I strongly suspect that changing `HGNAME` to `sl` instead of `hg` will break all of our internal stuff, so that's a non-starter.

bolinfest:
The other changes look safe, though!

kiilerix:
> I haven't tested this yet, but I strongly suspect that changing `HGNAME` to `sl` instead of `hg` will break all of our internal stuff, so that's a non-starter.

Right. I don't have enough information about how you use it in your setup. I dropped the last two changesets. They were not critical - just an attempt at cleaning things up.

quark-zju:
@kiilerix I modified the environment variable to `SAPLING_OFFLINE` and restores the `buildpyzip` logic. Let me know if you run into issues.

kiilerix:
Hi - Thanks for landing this.

It is surprising that the nice and clean history of the PR was squashed. I guess the PR was treated as what you seem to call a stacked change. I think it would be nice for acceptance by the wider the community that sapling both as tool and project also support the more standard workflow of just carefully using the tool as it kind of was intended.

quark-zju:
> It is surprising that the nice and clean history of the PR was squashed.

Sorry for the bad experience. This is currently limited by the [ImportIt](https://github.com/facebook/fbshipit/tree/main/src/importit) tool (maintained by another team). It does not support importing multiple commits as multiple internal code review units.

The Sapling GitHub repo is a mirror of a subset of the internal repo, synced by the [ShipIt](https://github.com/facebook/fbshipit) tool. Internally one code review unit maps to one commit. We don't use the merge/rebase features that GitHub provides because this repo is a mirror.

Sapling itself is not involved in merging a PR.

I do think importing commits as a stack is better in this case. But that requires building extra tools that do not exist right now. Let me open up an issue on the ShipIt side to see what they think.

kiilerix:
No problem. I'm just sharing the outside perspective. Thanks for considering it.

Also, interesting how ShipIt and sapling exposes the facebook-internal re-invention of https://arrenbrecht.ch/mercurial/pbranch/ ;-)

PR #454, opened 2023-01-14
fix windows install codeblocks

discentem:
This removes the `ps>` prefix in the windows installation instructions. With `ps>` included, you can't directly copy and past the commands into powershell


bolinfest:
Thanks, yes, this was bugging me, as well!

PR #459, opened 2023-01-14
tests: ignore __pycache__ directories created when running tests

strager:
tests: ignore __pycache__ directories created when running tests

### DIFF
--- a/eden/scm/tests/run-tests.py
+++ b/eden/scm/tests/run-tests.py
@@ -697,7 +697,7 @@ def parseargs(args, parser):
     if options.local:
         testdir = os.path.dirname(canonpath(sys.argv[0]))
         reporootdir = os.path.dirname(testdir)
-        pathandattrs = [("hg", "with_hg")]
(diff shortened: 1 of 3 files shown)


vegerot:
q: why not ignore all `pycache` directories?

strager:
> q: why not ignore all `pycache` directories?

I have no particular reason. Usually I prefer more specific entries in .gitignore.

eden/scm/tests/run-tests.py
```diff
@@ -697,7 +697,7 @@ def parseargs(args, parser):
     if options.local:
         testdir = os.path.dirname(canonpath(sys.argv[0]))
         reporootdir = os.path.dirname(testdir)
-        pathandattrs = [("hg", "with_hg")]
+        pathandattrs = [("sl", "with_hg")]
```
quark-zju, line 700: This would break internal build. Could you make it try both `sl` and `hg`?

quark-zju:
Closing since this was pushed as part of #469.

PR #461, opened 2023-01-14
phases: Delete dead code

strager:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/461).
* __->__ #461

phases: Delete dead code

Summary:
We check self._headbased twice, and the second time, we know that
self._headbased is true. Don't check twice; delete the dead code.

Test Plan:
Untested.


vegerot:
good catch

PR #456, opened 2023-01-14
smartlog: add option to hide '~' marker for unlogged ancestors

strager:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/456).
* __->__ #456
* #459
* #458

smartlog: add option to hide '~' marker for unlogged ancestors

Summary:
I made a template for 'sl smartlog' which is compact compared to the default:

    o  a0f0b95622       my commit
    o  10 hours ago     remote/main
    |
    ~

Unfortunately, the '|' and '~' lines take up a lot of space. Add an option to
disable these lines, making the output nice and tidy:

    o  a0f0b95622       my commit
    o  10 hours ago     remote/main

Test Plan:

    $ cd tests
    $ python run-tests.py test-log-graph-show-abbreviated-ancestors.t

PR #463, opened 2023-01-16
Allow `sl init --git` in non-empty directory

vegerot:
Allow `sl init --git` in non-empty directory

Summary: Currently, running `sl init --git` for a nonempty directory gives the
follwing error:
```
destination <..> already exists
```
This seems like an arbitary restriction to me, so I removed it.

It looks like this restriction was put in place because `sl init` and
`sl clone` share similar logic.  Is there any other reason for not allowing
`sl init` in non-empty directories?

Also, this commit fixes an imprecise error message when `init`ing a repository
in a non-directory (i.e. a file)

Test plan: Blocked by #447
manually tested with

(should pass, but didn't before)
```sh
DIR=/tmp/sl-clone-test zsh -c 'rm -rf $DIR; mkdir $DIR && cd $DIR && touch hi && CHGDISABLE=1 sl init --git .'
```

(should still fail)
```sh
DIR=/tmp/sl-clone-test REPO=git@github.com:97-things/97-things-every-programmer-should-know.git zsh -c 'rm -rf $DIR; mkdir $DIR && cd $DIR && CHGDISABLE=1 sl clone $REPO && CHGDISABLE=1 sl clone $REPO'
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/463).
* __->__ #463

### DIFF
--- a/eden/scm/edenscm/git.py
+++ b/eden/scm/edenscm/git.py
@@ -115,10 +115,10 @@ def clone(ui, url, destpath=None, update=True, pullnames=None):
 
     if os.path.lexists(destpath):
         if os.path.isdir(destpath):
-            if os.listdir(destpath):
+            if os.listdir(destpath) and url != "":
                 raise error.Abort(_("destination '%s' is not empty") % destpath)
         else:
-            raise error.Abort(_("destination '%s' already exists") % destpath)
+            raise error.Abort(_("destination '%s' exists and is not a directory") % destpath)
 
     try:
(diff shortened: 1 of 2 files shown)


strager:
Code looks good to me.

You should try writing an automated test. It's pretty easy. Just throw a file into the tests/ directory. See #456 for an example.

vegerot:
@strager great idea!  It was a lot easier than I thought <3

PR #466, opened 2023-01-16
tests: work around "Too many open files" failures

strager:


eden/scm/edenscm/testing/ext/hg.py
```diff
@@ -51,7 +51,7 @@ def testsetup(t: TestTmp):
             if setup:
                 testname = os.path.basename(testfile)
                 setup(testname, str(hgrcpath))
-    inprocesshg = True
```
quark-zju, line 54: This would make tests run much slower. We should only use it in incompatible tests.

quark-zju (changes requested):

quark-zju:
Feel free to reopen if you narrow it down to just related tests.

PR #473, opened 2023-01-18
website: arch linux installation instructions

lsjostro:
now easy to install sapling on Arch Linux distros with AUR enabled. 
https://aur.archlinux.org/packages/sapling-scm-bin


sggutier:
Nice! Thanks a lot for getting us into AUR and for updating the installation info @lsjostro !

sggutier (approved):

PR #494, opened 2023-01-20
(format): Align revsetlang elements table

vegerot:
(format): Align revsetlang elements table

Note: I also considered alingning the header line too, but then the second and
third columns were too spread apart

Test Plan: This commit does not change behavior

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/494).
* __->__ #494

### DIFF
--- a/eden/scm/edenscm/revsetlang.py
+++ b/eden/scm/edenscm/revsetlang.py
@@ -19,34 +19,36 @@ from .i18n import _
 from .pycompat import range
 
 
+# fmt: off
 elements = {
     # token-type: binding-strength, primary, prefix, infix, suffix
-    "(": (21, None, ("group", 1, ")"), ("func", 1, ")"), None),
-    "[": (21, None, None, ("subscript", 1, "]"), None),
-    "#": (21, None, None, ("relation", 21), None),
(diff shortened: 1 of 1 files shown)


bolinfest:
Although we don't have it wired up as a GitHub Action check (I'm in the process of reformatting our ghstack fork, which is tied to our internal license header linter, which makes it nontrivial...), but we have all of our Python code to be autoformatted using Black.

From https://stackoverflow.com/questions/58584413/black-formatter-ignore-specific-multi-line-code, it appears that there is a `# fmt: off` / `# fmt: on` escape hatch, though I don't know whether we want to grant that exception in this case. A quick `rg` didn't turn up uses of the escape hatch in our codebase.

vegerot:
@bolinfest thanks.  I added those formatter comments.

Let's land this diff or close it

quark-zju:
Our Python code is formatted automatically by [`black`](https://github.com/psf/black) and `usort`. So even if we take this change, it will be automatically reverted next day by automation like https://github.com/facebook/sapling/commit/a424ac8b96bfab832353cef4d93d888a60e1985b.

PR #495, opened 2023-01-23
[sl] honor NoDefaultCurrentDirectoryInExePath in NodeJs code

bolinfest:
[sl] honor NoDefaultCurrentDirectoryInExePath in NodeJs code

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/495).
* __->__ #495

### DIFF
--- a/addons/isl-server/proxy/startServer.ts
+++ b/addons/isl-server/proxy/startServer.ts
@@ -21,6 +21,7 @@ import crypto from 'crypto';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import {safeWhich} from 'shared/safeWhich';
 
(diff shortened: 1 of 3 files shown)


evangrayk (approved):
Looks good! We'll need to be careful to use this just about everywhere, maybe we should wrap `execa`/`child_process` or add a lint rule.

bolinfest:
This is in better shape now, but it still needs work. In particular, I need to add more tests.

quark-zju:
Is this still an issue? Suppose isl is launched from `sl` then the environment variable should be already set (inherited from sl), right?

PR #506, opened 2023-01-27
FreeBSD: Implement fstype

mshroyer:
FreeBSD: Implement fstype

Summary:
Adds an fstype implementation for FreeBSD's UFS and ZFS filesystems.  It is
uncommon, but possible, to configure ZFS filesystems to run in case-insensitive
mode, so we don't make assumptions about ZFS's case sensitivity.

Test Plan:
% cd eden/scm/lib/fsinfo/src
% ln -s ../examples/fstype.rs main.rs
% cd ..
% cargo run $PATH_TO_DIR_ON_UFS_OR_ZFS

### DIFF
--- a/addons/isl-server/src/analytics/serverSideTracker.ts
+++ b/addons/isl-server/src/analytics/serverSideTracker.ts
@@ -24,7 +24,7 @@ class ServerSideContext {
   }
 }
 
-const noOp = () => {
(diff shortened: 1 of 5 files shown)


mshroyer:
Closing because this stack didn't submit successfully (https://github.com/facebook/sapling/issues/511). Will re-export in a new stack.

PR #515, opened 2023-01-31
FreeBSD: Make run-isl.sh portable

mshroyer:
FreeBSD: Make run-isl.sh portable

Summary:
Remove dependency on bash-specific environment variable so ISL can start on
systems without it, such as FreeBSD.  We don't need to use $BASH_SOURCE[0]
instead of $0 because we're running the script rather than sourcing it.

Test Plan:
% gmake install-oss
% sl web

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/515).
* #516
* __->__ #515
* #514


evangrayk (approved):
thanks!

Issue #569, opened 2023-03-17
Feature: Allow for customization of commit template


steveluscher:
1. Since Sapling is most commonly used with GitHub, the commit template should use GitHub flavored markdown by default.
    ```
    <Replace this line with a title. Use 1 line only, 67 chars or less>

    ## Summary

    ## Test Plan
    ```
2. Beyond that, my request is for a way to customize the commit template, per repository.

steveluscher:
cc/ @discentem, @vegerot, who are working toward this in #524.

quark-zju:
#524 is about the PR body template, not commit template. The commit template is already customizable via configuration (but PR template is not). You might want to change [this line](https://github.com/facebook/sapling/blob/69c27551a0a705d0538b2c3a85a9f49e39355892/eden/scm/lib/config/loader/src/builtin_static/sapling.rs#L505) and surrounding configs to change the default template.

steveluscher:
Awesome! Is that documented anywhere? I looked and looked for it in the docs and couldn’t find anything.

opsengine:
@steveluscher see related PR https://github.com/facebook/sapling/pull/1209

PR #574, opened 2023-03-20
Use Markdown in commit template

steveluscher:
Commit message template in editor:

image

Rendering in GitHub PR:

image

Addresses #569.


fanzeyi:
Pass by comment:

I think this is going to be tricky since we have tons of automation running internally using `Summary: ` and `Test Plan: ` as a marker.

This PR is still reasonable though. I think we might want to consider shipping a different default for open source.

quark-zju:
How do you think about just leaving `emptymsg` empty? The "Summary" and "Test Plan" are Phabricator-ism that might not make sense everywhere.

PR #575, opened 2023-03-21
Update `githelp` to advise when `sl clean --dirs` is necessary

steveluscher:
Whenever there's a path or the `-d` flag, `git clean` implies that cleaning should be done recursively, into directories, removing empty ones as it goes.

The equivalent in Sapling is `sl clean --dirs`.

Closes #564.

PR #576, opened 2023-03-21
Add `m4` to Ubuntu requirements doc

steveluscher:
I don't know what this is, but I found that I needed it to run `./build.sh`.


bolinfest:
What is this file used for: is this just for EdenFS and not the Sapling client?

steveluscher:
By ‘the file’ do you mean `requirements_ubuntu.txt`? I believe it’s to make our lives easier here, on the outside, by telling us what we need to `apt-get install` if we hope to be able to build the project.

bolinfest:
Yes, just confirming what you're trying to build since, at least for the client, the de facto list is what we use when we configure the `Dockerfile` for the builds:

https://github.com/facebook/sapling/blob/d1d5134f707f1427b0f6f9750a339ed879791a28/.github/workflows/sapling-cli-ubuntu-20.04.Dockerfile#L17

https://github.com/facebook/sapling/blob/d1d5134f707f1427b0f6f9750a339ed879791a28/.github/workflows/sapling-cli-ubuntu-22.04.Dockerfile#L17

The nice thing about the `Dockerfile` is that it is "proof" of the minimal set of dependencies required to build since it starts from a barebones Ubuntu image.

steveluscher:
I suppose that @genevievehelsel just left this file here, 3 years ago, to be helpful – which I found it to be – because I couldn't run `./build.sh` until I had at least that stuff installed.

Maybe you should just delete it.

bolinfest:
@steveluscher again, which part of the project are you trying to build? Admittedly, we don't have a `Dockerfile` and whatnot for building EdenFS, so perhaps this is still necessary?

Unfortunately, we've only had the time to tighten up the client build process, docs, etc. so far.

steveluscher:
I’m not trying to be obtuse. The beginning and the end of what my goal is, is to ‘run `./build.sh` without errors.’

Think about it from the perspective of a first time contributor. Imagine they’re asked to submit a PR to fix `githelp`, and they do (#575).

This person has no idea

* how to test what they’ve done
* how to run the tests they’ve written
* how to produce a build from the code they changed
* how to get permission to build their PR in CI

…and so they find `./build.sh` and run it, hoping that something good will happen; hoping that it will produce a result similar to what the CI would produce.

Then, when that script fails, they nevertheless persist, finding and installing missing dependencies.

Then they open this PR, in hopes that these findings will help the next person.

That’s all.

zzl0:
Hi @steveluscher Thanks for your interest in contributing to Sapling and sorry for the confusion about the doc and code. I believe the `build.sh` and `requirements_ubuntu.txt` are for building EdenFS.

For building Sapling client (`sl`) from source code, you can check the instructions on our website: https://sapling-scm.com/docs/introduction/installation#building-from-source

And if you have any questions, you can also join our Discord to discuss them there. You can find the Discord at the bottom of the our website https://sapling-scm.com/

With that said, I will improve our README to avoid this kind of confusion.

PR #596, opened 2023-04-07
fix subcommands in pr.md

skevy:
The markdown was broken for this file.


zzl0:
The markdown were generated by a script. How about updating the script here? https://github.com/facebook/sapling/blob/main/website/scripts/generate-command-markdown.py#L283

zzl0:
I will import it and update it accordingly, thanks

zzl0 (approved):
Thanks for the PR. There is minor issue, I will fix it.

PR #604, opened 2023-04-17
Update Repository.ts

tomasreimers:
Github also allows repos of the form `github.com:owner/repo.git`

This updates the regex to support those.


evangrayk:
Thanks for the PR! would you mind adding a unit test for this as well in `isl-server/src/__tests__/Repository.test.ts` (under `extractRepoInfoFromUrl`)?

Issue #304, opened 2022-12-07
Cannot build reviewstack


viveksjain:
For various reasons I am trying to self-host an instance of reviewstack. To do so I'm following https://github.com/facebook/sapling/blob/main/addons/reviewstack/README.md. I have run `yarn` in the addons/ folder to install all the dependencies. The next step of `yarn codegen` in addons/reviewstack fails:
```
sh$ yarn codegen
yarn run v1.22.19
$ node ./codegen.js
node:internal/errors:484
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/vivek.jain/sapling/addons/textmate/dist/extensions' imported from /Users/vivek.jain/sapling/addons/textmate/dist/index.js
    at new NodeError (node:internal/errors:393:5)
    at finalizeResolution (node:internal/modules/esm/resolve:260:11)
    at moduleResolve (node:internal/modules/esm/resolve:879:10)
    at defaultResolve (node:internal/modules/esm/resolve:1087:11)
    at nextResolve (node:internal/modules/esm/loader:161:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:831:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:413:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:77:40)
    at link (node:internal/modules/esm/module_job:76:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v19.2.0
node:child_process:960
    throw err;
    ^

Error: Command failed: node --experimental-specifier-resolution=node ../textmate/dist/index.js src/generated/textmate/TextMateGrammarManifest.ts ../reviewstack.dev/public/generated/textmate
node:internal/errors:484
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/vivek.jain/sapling/addons/textmate/dist/extensions' imported from /Users/vivek.jain/sapling/addons/textmate/dist/index.js
    at new NodeError (node:internal/errors:393:5)
    at finalizeResolution (node:internal/modules/esm/resolve:260:11)
    at moduleResolve (node:internal/modules/esm/resolve:879:10)
    at defaultResolve (node:internal/modules/esm/resolve:1087:11)
    at nextResolve (node:internal/modules/esm/loader:161:28)
[...]

PR #613, opened 2023-04-22
Fixing textmate grammar build issue for node 19 when trying to build reviewstack locally.

Flux159:
**Summary:**
In https://github.com/facebook/sapling/issues/304 it was mentioned that reviewstack builds fail on Node 19 because `--experimental-specifier-resolution=node` is no longer supported in Node 19.

This diff uses the `--experimental-loaders` api available in v19 and adds a loader to allow for the textmate grammars build to complete properly.

**Test Plan:**
I was able to follow the local development guide at https://github.com/facebook/sapling/tree/main/addons/reviewstack#local-development and get reviewstack running locally with Node 19.

Specifically, before `yarn codegen` in the `addons/reviewstack` directory would fail trying to build textmate grammars w/ v19. Now it completes successfully.


addons/reviewstack/loader.mjs
```diff
@@ -0,0 +1,21 @@
+// Found here: https://stackoverflow.com/a/74931879
+import {existsSync} from 'fs';
```
bolinfest, line 2: Hi Suyog, good to hear from you!

Would you mind running this through Prettier or something so we terminate lines with `;` where appropriate, prefer `const` to `let`, etc.?

addons/reviewstack/loader.mjs
```diff
@@ -0,0 +1,21 @@
...
+import {basename, dirname, extname, join} from 'path'
+import {fileURLToPath} from 'url'
+
+let extensions = ['mjs', 'js', 'json'], resolveDirs = false
+
+let indexFiles = resolveDirs ? extensions.map(e => `index.${e}`) : []
+let postfixes = extensions.map(e => `.${e}`).concat(indexFiles.map(p => `/${p}`))
+let findPostfix = (specifier, context) => (specifier.endsWith('/') ? indexFiles : postfixes).find(p =>
```
bolinfest, line 10: This doesn't seem like it will work on Windows?

Flux159: Just realized that same Stackoverflow answer this is from made a package that works on windows too - https://www.npmjs.com/package/extensionless ; still same test plan of running. 

`yarn` in `/addons` folder
`yarn codegen` in `addons/reviewstack`
`yarn start`

Note that `yarn codegen` is running `addons/reviewstack/textmate.js` which also doesn't look like it will run on windows since using unix style paths.

quark-zju:
The `--experimental-specifier-resolution=node` problem is now fixed in the main branch (https://github.com/facebook/sapling/commit/f7d3b814850389316c7b0840855eaaf6f185cff0).

Issue #632, opened 2023-05-26
Allow customizing commit template for ISL


alex-statsig:
Sapling CLI  uses the `committemplate.changeset` config as the default commit message when using `sl commit`. However, ISL doesn't pull this in. It seems like it tries to run a command `sl debugcommitmessage` to get the default message, which fails (I'm guessing that's available in mercurial internally). It seems like changing this to use a config would be better (unless its supposed to be based on changes in the commit, but I didn't see it get called a second time).

I tried adding an alias for `debugcommitmessage = !echo "\n# Summary ..."`, but unfortunately somehow it seems like the aliases aren't used (I still got "unknown command 'debugcommitmessage' (use 'sl help' to get help)" when running ISL locally).

PR #633, opened 2023-05-26
Allow config for custom ISL commit message template

alex-statsig:
Allow config for custom ISL commit message template

# Summary
Regarding https://github.com/facebook/sapling/issues/632, allows a sapling config to be set which gives a default template for ISL title/description.

Currently namespaced under `committemplate.isl` (with other committemplate configs), but could flip this to `isl.committemplate` (with other isl configs).

Also could see this being set by a version controlled file in the repo to standardize by repo instead of sapling config (ex. `<root>/.sapling-isl-committemplate`), open to thoughts on that

This is necessary because debugcommitmessage isn't a public command, so this could potentially be fixed by adding that to sapling CLI instead.

# Test Plan
Templated this commit using it, after running `sl config -l committemplate.isl "<title># Summary\n\n# Test Plan\n\n"`


quark-zju (changes requested):
This does not seem like the right approach. The configs under the `[committemplate]` are all template expandable. So if you have something like:

    [committemplate]
    isl = {title} ...
    title = my-title-definition ...

The `{title}` should be expanded, not returned as is.
 
How do you think about:
- Move the `debugcommitmessage` extension to `commands/debug.py`. Sunset the extension.
- Make ISL use `debugcommitmessage isl` so if you want to use a different template from CLI, you can define the `changeset.isl` config under `[committemplate]`.

alex-statsig:
> This does not seem like the right approach. The configs under the `[committemplate]` are all template expandable. So if you have something like:
> 
> ```
> [committemplate]
> isl = {title} ...
> title = my-title-definition ...
> ```
> 
> The `{title}` should be expanded, not returned as is.
That makes sense, I wasn't sure if the commit info was setup to handle a context-based default (such as including title / files changed etc.). I considered using a command like `!(HGEDITOR="cat" sl commit 2>/dev/null || true)` to extract the expanded version.

> How do you think about:
> 
> * Move the `debugcommitmessage` extension to `commands/debug.py`. Sunset the extension.
I'm not sure I follow the purpose / effect of this

> * Make ISL use `debugcommitmessage isl` so if you want to use a different template from CLI, you can define the `changeset.isl` config under `[committemplate]`.
Actually my goal wasn't to distinguish it from CLI (although I think it is a plus), my goal was just to be able to set a template. So perhaps the real issue here is that the template is not currently working in ISL. I see this log in extension output:
```
Could not fetch commit message template Error: Command failed with exit code 255: sl debugcommitmessage --noninteractive
unknown command 'debugcommitmessage'
(use 'sl help' to get help)
    at command (/Users/alex/dev/sapling/addons/node_modules/execa/lib/error.js:66:2)
    at /Users/alex/dev/sapling/addons/node_modules/execa/index.js:125:5
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```
I tried adding my own version via an alias `debugcommitmessage = !(HGEDITOR="cat" sl commit 2>/dev/null || true)` but this wasn't picked up by the extension still. I assumed `debugcommitmessage` was only available internally or something.

quark-zju:
> `debugcommitmessage`

It is provided by the extension called "debugcommitmessage". You can use `--config extensions.debugcommitmessage=` to enable it. To make it always available, you can move [the command from the extension](https://github.com/facebook/sapling/blob/main/eden/scm/edenscm/ext/debugcommitmessage.py#L14-L45) to [builtin commands](https://github.com/facebook/sapling/blob/main/eden/scm/edenscm/commands/debug.py), then add "debugcommitmessage" to [the ignored extensions list](https://github.com/facebook/sapling/blob/main/eden/scm/edenscm/extensions.py#L35).

alex-statsig:
> > `debugcommitmessage`
> 
> It is provided by the extension called "debugcommitmessage". You can use `--config extensions.debugcommitmessage=` to enable it. To make it always available, you can move [the command from the extension](https://github.com/facebook/sapling/blob/main/eden/scm/edenscm/ext/debugcommitmessage.py#L14-L45) to [builtin commands](https://github.com/facebook/sapling/blob/main/eden/scm/edenscm/commands/debug.py), then add "debugcommitmessage" to [the ignored extensions list](https://github.com/facebook/sapling/blob/main/eden/scm/edenscm/extensions.py#L35).

Got it, makes sense. Thanks for the explanation, I can update the PR to do that then

alex-statsig:
@quark-zju Updated, confirmed `sl debugcommitmessage (isl)` works without enabling the extension now

alex-statsig:
@quark-zju is there a way to version control the `<repo>/.sl/config` file or would you recommend something external like a "prepare" script which runs `sl config -l committemplate.changeset.isl "..."`? Trying to find the best way to utilize this template to standardize our summaries now

PR #651, opened 2023-06-14
Fix github check status logic

alex-statsig:
Fix github check status logic

# Summary
The current logic for github checks uses "checkSuites" to get all the check suites run for the most recent commit. This doesn't actually match which checks are shown overall for the PR (ex. some may be duplicated and cancelled, and thus not be important, or may not be blocking).

In particular, I noticed for my repo that there is some QUEUED "checkSuites" which don't appear in the github UI, and there are some cancelled checkSuites which were superceded by a later run, and thus are acceptable to fail. This leads to passing PRs showing as "pending", and some PRs showing as "failing" if a run was cancelled and rerun for the latest commit.

This PR updates to use `commit.statusCheckRollup.state` instead of manually trying to aggregate the check suites. If the breakdown is desired, `commit.statusCheckRollup.contexts.checkRunCountsByState` could provide that (but the aggregated "state" seems to handle it well).

# Test Plan
This seems to give accurate results within my repo (tested with some pending, passing, and failing cases)

### DIFF
--- a/addons/isl-server/src/CodeReviewProvider.ts
+++ b/addons/isl-server/src/CodeReviewProvider.ts
@@ -19,6 +19,8 @@ type DiffSummaries = Map<DiffId, DiffSummary>;
  * API to fetch data from Remote Code Review system, like GitHub and Phabricator.
  */
 export interface CodeReviewProvider {
+  getDiffUrl(diffId: DiffId): string;
+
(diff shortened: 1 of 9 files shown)


evangrayk:
nice, this looks good! thank you for contributing!

PR #684, opened 2023-07-23
reviewstack: fix github stacked diff display

ahornby:
github diffs were showing the full set of commits in the stack rather than just the commits for the given PR version

gitHubPullRequestComparableVersions was returning null for base commit,  which meant we always fell back on the } else if (afterBaseCommitID != null) { branch of gitHubPullRequestVersionDiff.  That fallback used to work fine, but now its giving us the base of the full stack rather than the base of the PR. I don't know how it worked before (perhaps github api behaviour changed?), but there you go!

Fix is to use the version.baseParent in gitHubPullRequestComparableVersions, which has the right value in it.

Test Plan:

yarn test --watchAll=false

Local build and run:

Before, https://reviewstack.dev/facebookincubator/fizz/pull/96 shows changes for full stack:
image

After, http://localhost:3000/facebookincubator/fizz/pull/96 shows only relevant changes for PR:
image

### DIFF
--- a/addons/reviewstack/src/recoil.ts
+++ b/addons/reviewstack/src/recoil.ts
@@ -852,11 +852,11 @@ export const gitHubPullRequestComparableVersions = atom<ComparableVersions>({
       const [versions, selectedVersionIndex] = get(
         waitForAll([gitHubPullRequestVersions, gitHubPullRequestSelectedVersionIndex]),
       );
-      const latestCommit = versions[selectedVersionIndex].headCommit;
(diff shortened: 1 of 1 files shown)


alex-statsig:
Is there a timeline for reviewstack deployments? Curious when we can expect to see this fix, since it's causing reviews to be pretty hard atm

sggutier:
We don't really have one yet, but I just pushed a new deploy. That should include this change.

Issue #652, opened 2023-06-14
[ISL][Feature Request] Show diff review status in ISL


alex-statsig:
Currently "diff status" (OPEN/CLOSED/DRAFT) is shown in the ISL, along with # of comments and checks status (passing/failing checks). Another very important thing to show to indicate if a PR is ready to merge or needs updating is the review status ("Accepted", "Changes Requested", "Review Required"). I know a similar concept exists within Meta's phabricator, and I think it may already be shown in the internal ISL.

This could potentially be a new badge alongside diff/checks status , or it could be a new state of `DiffBadgeContent` (i.e. if the diff is open, there is a sub-state of "Accepted" vs "Changes Requested" etc.).

I would be happy to contribute a PR adding this, if any guidance could be given regarding how it is handled for the internal diff provider (to avoid forking things).

Here's a rough example I threw together locally:
<img width="363" alt="Screen Shot 2023-06-14 at 12 25 33 AM" src="https://github.com/facebook/sapling/assets/77301670/1e7b067f-544e-43fc-9868-9217649873fb">

alex-statsig:
Looks like this was just rolled out via #698 , thanks so much @evangrayk !

PR #698, opened 2023-08-01
[isl] Add badge for GitHub PR review decision

evangrayk:
[isl] Add badge for GitHub PR review decision

## Summary:

This diff relates to [Issue #652](https://github.com/facebook/sapling/issues/652).

It's useful to know the review state of a PR. We do this for diffs internally, and in fact that's the main "status" we show. GitHub differentiates between "Open/Closed/Merged" but separately differentiates independent review states "Approved/Changes Requested/Review Required".

We should show this review state as well as status.

I considered merging the two, so it's more like we have internally. But I think it makes more sense to keep this separate, to be true to GitHub's data model.

## Test Plan: 

Screenshot 2023-08-01 at 11 33 44 AM
Screenshot 2023-08-01 at 11 33 55 AM

I tried a couple of other UI variants, including one where the review states are not badge but just text. This would have matched GitHub's design more closely. But I couldn't make it fit quite right. I felt like the badges are easier to read at a glance. Happy to get feedback about this though.

### DIFF
--- a/addons/isl-server/src/github/generated/graphql.ts
+++ b/addons/isl-server/src/github/generated/graphql.ts
@@ -24257,7 +24257,7 @@ export type YourPullRequestsQueryVariables = Exact<{
 }>;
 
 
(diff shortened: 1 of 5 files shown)


sggutier (approved):
LGTM!

PR #769, opened 2023-11-10
fix mononoke build

ahornby:
fix mononoke build

Makes usage of anyhow::Error::backtrace() conditional on fbcode_build. This is necessary because the cargo build of anyhow disables its backtrace features when using RUSTC_BOOTSTRAP=1

For oss cargo build to work on current rustc stable 1.73+ I hid the #![feature(provide_any)] usage by making it condition on fbcode_build.

Test plan:

local build with `./build/fbcode_builder/getdeps.py --allow-system-packages build --src-dir=. mononoke`

github CI, done by regenerating the mononoke github actions with: `./build/fbcode_builder/getdeps.py --allow-system-packages generate-github-actions --free-up-disk --src-dir=. --output-dir=.github/workflows --job-name="Mononoke " --job-file-prefix=mononoke_ mononoke`

Before, broken

After, works

### DIFF
--- a/.github/workflows/mononoke_linux.yml
+++ b/.github/workflows/mononoke_linux.yml
@@ -10,11 +10,14 @@ on:
     branches:
     - main
 
+permissions:
+  contents: read  #  to fetch code (actions/checkout)
+
 jobs:
   build:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Show disk space at start
       run: df -h
     - name: Free up disk space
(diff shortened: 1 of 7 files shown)


xavierd:
> For build to work on current rustc stable 1.73+ I had to remove the #![feature(provide_any)] usage. You may need to be on rustc 1.73+ internally to import this successfully

It looks like our internal build is unfortunately failing to build. The compiler is sending us to https://github.com/rust-lang/rust/issues/96024 which appears to still be open. Can you re-send this PR while keeping the feature?

ahornby:
> It looks like our internal build is unfortunately failing to build. The compiler is sending us to [rust-lang/rust#96024](https://github.com/rust-lang/rust/issues/96024) which appears to still be open. Can you re-send this PR while keeping the feature?

done, added in cfg_attr usage to keep it off in oss build but on otherwise

PR #427, opened 2023-01-05
add config option to disable mention of ReviewStack in PR bodies

discentem:
add config option to disable mention of ReviewStack in PR bodies

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/427).
* __->__ #427

PR #779, opened 2023-11-22
[pr body update] fix reviewstack option breaking stack list detection

vuduchild:
In #427 an option was added to remove the ReviewStack link from the PR stack list in the PR body. However, removal of that line entirely breaks [parse_stack_information](https://github.com/facebook/sapling/blob/1e4d8eca58f5096b25dd93fc31bfa4c0667ce343/eden/scm/sapling/ext/github/pull_request_body.py#L93) as it expects to find a line beginning with "Stack created with [Sapling]".

To fix this, I've added a new textual marker (`[//]: # (BEGIN SAPLING FOOTER)`) which markdown doesn't render. From now on, to detect a stack, we'll search for this marker and then this should allow us to split on it. For backward compatibility, I'm leaving in the previous marker detection as well.

Tests: submitted this PR with the proper stack listing :) Also, take a look at downstream PR #780 that also relies on this working.

---
* #780
* __->__ #779

### DIFF
--- a/eden/scm/sapling/ext/github/mock_utils.py
+++ b/eden/scm/sapling/ext/github/mock_utils.py
@@ -297,6 +297,7 @@ class MockGitHubServer:
             ]
             body += (
                 "\n---\n"
+                "[//]: # (BEGIN SAPLING FOOTER)\n"
                 "Stack created with [Sapling](https://sapling-scm.com). Best reviewed"
                 f" with [ReviewStack](https://reviewstack.dev/{owner}/{name}/pull/{pr_number}).\n"
                 + "\n".join(pr_list)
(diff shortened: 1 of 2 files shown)


vuduchild:
@rejc2 - indeed a more elegant solution. I'll change it.

vuduchild:
@rejc2 - after further examination, I've realized that for backward compatibility we'll have to use either this new markdown-invisible marker or the old `Created with Sapling` marker.

quark-zju:
There are failing tests:
- test-ext-github-pr-submit-overlap.t
- test-doctest.py

Could you try to run the tests (via run-tests.py after build a local `sl`) and see if you can reproduce issues?

vuduchild:
@quark-zju - I fixed the tests (one of them actually revealed a bug. Yay!)

keunhong:
Hello, this seems to break ReviewStack. Deleting the footer makes ReviewStack detect the stack properly again. Related: https://github.com/facebook/sapling/issues/826

quark-zju:
This seems to be pushed as part of the next PR.

PR #780, opened 2023-11-22
[pr body update] update stack list without overwriting PR title and body

vuduchild:
The current behavior of PR updates is flawed - every time one submits a new version of a PR, it is followed by an update of the title and description body of the PR. Often, the PR author would edit the PR title in Github's web UI after submission, to find their marvelous prose offhandedly deleted by Sapling.

In this diff I'm introducing a new config option (`github.preserve-pull-request-description`) to control this overwrite behavior. With the option turned on, subsequent updates **only overwrite the stack listing** in the PR description to reflect PR stack changes, while the actual PR description and title are left alone.

Here's the new behavior in detail:
- A new PR is submitted using `sl pr submit`
  - The title and description are created out of the commit message.
- The author updates the title and description in the Github web UI
- The author makes changes in their local repo, amends and and runs `sl pr submit` again
  - The PR title is left alone.
  - The PR description is left alone, except for the stack list at the bottom which gets updated

At this point, I've set the option to default to `False` to avoid surprising current users, preserving current (albeit flawed) behavior. However, I believe over time this should default to True. The reasoning being that technically the local commit and remote PR should be 2 separate entities (a PR may contain more than one commit etc).

Test:
- Submit a PR
- Edit the title and description of the PR to make it different than the default commit message content.
- Create another PR stacked on top of the first one
- Submit the new PR
- Observe the old PR now has a stack list
- Amend either of the PRs locally with a new change
- Submit an update to the stack
- Verify the PR titles and descriptions have not changed.

**NOTE**: This PR is dependant on downstream #779 which fixes a bug affecting the parsing and detection of the stack list in the PR description. 

---
* __->__ #780
* #779

### DIFF
--- a/eden/scm/sapling/ext/github/__init__.py
+++ b/eden/scm/sapling/ext/github/__init__.py
@@ -47,6 +47,7 @@ configtable = {}
 configitem = registrar.configitem(configtable)
 
 configitem("github", "pull-request-include-reviewstack", default=True)
+configitem("github", "preserve-pull-request-description", default=False)
 
 
 @hint("unlink-closed-pr")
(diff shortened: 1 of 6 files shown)


amyworrall:
I really want this behaviour! Looking forward to the PR landing.

PR #781, opened 2023-11-22
[isl] increase width of diff window in split stack edit panel

vuduchild:
The width of the a diff window in the stack splitting dialog in ISL is too narrow and isn't adaptive to screen sizes. It makes it unnecessarily hard to split code formatted with longer lines when working on larger screens and resolutions.

Tested with a longer line length on a small and large screen.

---
Stack created with [Sapling](https://sapling-scm.com).
* #782
* __->__ #781


vuduchild:
@evangrayk that would work very nicely! I wish I thought of that. I'll update and publish the PR.

vuduchild:
Thanks @evangrayk !

PR #782, opened 2023-11-22
[isl] reduce number of pr fetched in graphql query to avoid timeouts

vuduchild:
When using 100 PRs, I often encounter errors resulting of Github timeouts. Reducing the number of PRs we're fetching solves the issue.
A possibly more elegant future solution would be to specify which PRs' data we're interested in fetching.

---
Stack created with [Sapling](https://sapling-scm.com).
* __->__ #782
* #781


evangrayk:
Thanks for the PR! 

20 is a little low, I would certainly start running into that limit in my work. But if we're currently unable to fetch because the limit is too high we probably should reduce it. Not sure what the right number is honestly. We probably need to be doing some batching somehow to support fetching however PRs many we need, but not all in one go to avoid errors.

Does your problem get fixed even with a higher number? say, 50?

vuduchild:
@evangrayk - yes, 50 still works, just makes the request take slightly longer. I'll change it.

My suggestion for a future implementation would be to still batch the call, but to perhaps experiment with specifying which PRs we're interested in (if the API supports it). This should limit the size of the results and perhaps reduce the server processing time.

PR #801, opened 2024-01-01
[reviewstack] Updating textmate build to append relative import .js prefix

Ahmed-Ali:
[reviewstack] Updating textmate build to append relative import .js prefix

With more current versions of node, the `--experimental-specifier-resolution=node'` does not work anymore and running `yarn codegen` from the `reviewstack` folder will fail because of textmate module.

This PR adds a step in the `addons/reviewstack/textmate.js` to fix textmate imports to use .js extension. This will only impact the transpiled .js files and not the source .ts files


quark-zju:
Sorry for the late reply - I didn't fully understand the context. We generally prefer less dependencies, like, no `fix-esm-import-path`. I found a fix without the dep: https://github.com/facebook/sapling/commit/f7d3b814850389316c7b0840855eaaf6f185cff0.

This should now be fixed in the main branch. Thanks for your PR!

PR #804, opened 2024-01-02
[CI] shrink builder-images by 60% / several GB

shish:
[CI] shrink builder-images by 60% / several GB

If we copy the repo into docker, save the state, then delete the repo, then the repo is still there in the history

These changes use docker staged builds - having a "base" stage which installs the tools needed, a "populate-cache" stage which uses ~3GB of temporary disk space to populate the 162MB cache, and then a "main" stage which copies the 162MB cache but leaves the rest of the temporary files behind

Before - around 5GB:
```
$ docker build -t sapling-build . -f .github/workflows/sapling-cli-ubuntu-22.04.Dockerfile
$ docker history sapling-build
IMAGE          CREATED              CREATED BY                                      SIZE      COMMENT
33b251176a7d   26 seconds ago       RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   0B        buildkit.dockerfile.v0
<missing>      28 seconds ago       WORKDIR /root                                   0B        buildkit.dockerfile.v0
<missing>      28 seconds ago       RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   0B        buildkit.dockerfile.v0
<missing>      28 seconds ago       RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   1.56GB    buildkit.dockerfile.v0
<missing>      About a minute ago   RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   2.19MB    buildkit.dockerfile.v0
<missing>      About a minute ago   RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   8.97MB    buildkit.dockerfile.v0
<missing>      About a minute ago   WORKDIR /tmp/repo                               0B        buildkit.dockerfile.v0
<missing>      About a minute ago   COPY . /tmp/repo # buildkit                     1.17GB    buildkit.dockerfile.v0
<missing>      About a minute ago   ENV PATH=/root/.cargo/bin:/usr/local/sbin:/u…   0B        buildkit.dockerfile.v0
<missing>      About a minute ago   RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   1.49GB    buildkit.dockerfile.v0
<missing>      2 minutes ago        RUN |1 DEBIAN_FRONTEND=noninteractive /bin/s…   421MB     buildkit.dockerfile.v0
[...]

PR #806, opened 2024-01-04
Update build-images to allow running sl as well as building it

shish:
Update build-images to allow running sl as well as building it

My laptop has weird company-customised versions of various tools which makes it a pain to build vanilla OSS projects, so I'd like to do development work inside a container. Using the CI build-images seems smarter than trying to come up with my own DIY dev-images, the only downside being that there are a couple of missing runtime-dependencies. Eg, when running ./sl, I get an error that the locale is invalid:

```
root@e11f03ad868a:/workspaces/sapling/eden/scm# ./sl
Traceback (most recent call last):
  File "/workspaces/sapling/eden/scm/sapling/__init__.py", line 64, in run
    from . import dispatch
  File "/workspaces/sapling/eden/scm/sapling/dispatch.py", line 27, in <module>
    from . import (
  File "/workspaces/sapling/eden/scm/sapling/cmdutil.py", line 29, in <module>
    from . import (
  File "/workspaces/sapling/eden/scm/sapling/crecord.py", line 32, in <module>
    locale.setlocale(locale.LC_ALL, "")
  File "/usr/lib/python3.10/locale.py", line 620, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/806).
* #807
* __->__ #806
* #804


quark-zju:
This seems to be included in #807. We don't have good tooling to handle stacked PRs yet but we hope to improve in the future.

PR #807, opened 2024-01-06
Add a devcontainer config

shish:
Add a devcontainer config

My laptop is full of weird versions of weird tools. Sapling requires specific versions of specific tools. It's nice to be able to open the sapling folder in vscode and get a prompt for "would you like to do your development inside a container with all the correct tools pre-installed?" :) This also allows people on github to click a button to open in github's web-based container-IDE and start developing without installing anything at all, though I don't know if we'd expect anybody to do that

For starters I'm using the CI build image without changes to try and keep the maintenance burden as small as possible, though it would be possible to use a separate layer within the Dockerfile (eg if we wanted to install extra tools that are useful for dev but not needed for building), or have a totally separate Dockerfile


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,24 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
```
quark-zju, line 2: These broke our internal linter about "valid JSON". So I moved it to a separate file.

PR #815, opened 2024-01-18
doc: fix GitHub PR link in `getting-started.md`

pseudomata:
The 'Getting Started' documentation page has a section about replacing the GitHub PR URL with reviewstack, but the link in Markdown is parsed as GitHub flavored Markdown and shows just `facebook/react#25506` instead of `https://github.com/facebook/react/pull/25506` which is slightly confusing. This change fixes that to the full intended URL is displayed.

See: https://sapling-scm.com/docs/introduction/getting-started#browsing-pull-requests
image

### DIFF
--- a/website/docs/introduction/getting-started.md
+++ b/website/docs/introduction/getting-started.md
@@ -170,7 +170,7 @@ In addition to reviewing pull requests on github.com, you may also want to try [
 
 To view a GitHub pull request on ReviewStack, take the original URL:
 
-> https://github.com/facebook/react/pull/25506
+> [https://github.com/facebook/react/pull/25506](https://github.com/facebook/react/pull/25506)
 
 and replace the `github.com` domain with `reviewstack.dev`:
 


quark-zju:
Thanks! I removed the code span so they render as regular text.

pseudomata:
In my first commit I didn't wrap them in a code span but they still appeared to render as `org/project#1` instead of `github.com/org/project/pull/1` (at least in the GitHub diff for the rendered Markdown). Since [Docusaurus uses GitHub flavored Markdown](https://v1.docusaurus.io/docs/en/doc-markdown) I suspect it would be the same without the code span.

quark-zju:
> they still appeared to render as `org/project#1` instead of `github.com/org/project/pull/1`

I noticed that. I found appending a space also works so I used that instead.

quark-zju:
I think this was pushed as https://github.com/facebook/sapling/commit/fe03a1760d26d2afb594db7c6d6fb804d8ee2a4d. Not sure why the bot didn't close this...

pseudomata:
Ah cool. Thank you @quark-zju!

PR #816, opened 2024-01-18
Add "open in code review" context menu action

alex-statsig:
Add "open in code review" context menu action

A common feature from other SCM extensions is a context-menu item to copy a link / open the selection/file in the remote SCM provider (ex. open in Github). This adds support for that in the editor / file explorer.

Couple of things I wasn't totally confident about that I'd like thoughts on:
- I opted to just add this to context menu in vscode since that's where most other extensions do this, but I'm guessing this might be desired in ISL too, which might changed some implementation details
- Open to other titles for the commands, but afaict the message can't be dynamic (i.e. use code review provider's name), so I settled on "Open in Code Review"
- I went with the simple approach of copying a link to the file on main using the same line numbers. This won't necessarily work if the file is dirty or modified in remote main. This could be mitigated by using the most recent public commit instead of main, and perhaps some magic with blame, but seemed more complex
- Not sure the best way to grab the remote branch's name; just hard-coded "main" in my testing, but it could be "master" or something else as well. Is this easily retrievable from sapling?
- I started setting a context for "sapling:hasRepo" so the option would only show if the editor has some folder with a sapling repo. I could be smarter and check if the active file is in a repo, but this seemed effective enough and simpler (scmProvider used in other commands is only available with the SCM panel)

### DIFF
--- a/addons/isl-server/src/CodeReviewProvider.ts
+++ b/addons/isl-server/src/CodeReviewProvider.ts
@@ -45,6 +45,13 @@ export interface CodeReviewProvider {
   getDiffUrlMarkdown(diffId: DiffId): string;
   getCommitHashUrlMarkdown(hash: string): string;
 
+  getRemoteFileURL?(
+    path: string,
+    publicCommitHash: string | null,
+    selectionStart?: {line: number; char: number},
+    selectionEnd?: {line: number; char: number},
+  ): string;
+
   updateDiffMessage?(diffId: DiffId, newTitle: string, newDescription: string): Promise<void>;
 
   getSuggestedReviewers?(context: {paths: Array<string>}): Promise<Array<string>>;
--- a/addons/isl-server/src/Repository.ts
+++ b/addons/isl-server/src/Repository.ts
@@ -677,9 +677,14 @@ export class Repository {
     }
   });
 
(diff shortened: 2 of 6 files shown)


alex-statsig:
@quark-zju / @evangrayk could you take a look at reviewing this and https://github.com/facebook/sapling/pull/717 when you get a chance? These are the main complaints I get about migration from git -> sapling still, and would love to close these out

alex-statsig:
@quark-zju anything I can do to help get this merged faster (ex. put it behind a setting)? Would love to get this feature in so my company can start using it to share code links more easily (along with a few other PRs like https://github.com/facebook/sapling/pull/827 and https://github.com/facebook/sapling/pull/717)

quark-zju:
Sorry for the late reply. I wasn't very familiar with the vscode extension code base. It seems we do have a similar feature internally that is provided by a different vscode extension.

- Could `getRemoteFileURL` take a commit hash? Branches like `main` are moving targets. A commit hash would work better. If the commit hash is hard to get, it seems you can make it default to `HEAD` (but it still seems nice to be able to provide the hash).
- `sapling:hasRepo` seems like it will cause false positives. Could it be changed to test if the repo has a GitHub code review provider attached or not?

PR #818, opened 2024-01-20
Don't force en_US.utf-8 as locale as it mgiht not exist.

### LINKED ISSUE #814: VSCode/ISL integration sometimes shows "Not a valid repository" on WSL
I'm not sure how to gather more debug information, but the VSCode extension appears to sometimes not detect correctly that the open folder *is* a valid Sapling checkout:

> Not a valid repository
> /home/flameeyes/repos/unpaper is not a valid Sapling repository. Clone or init a repository to use ISL.

```
(venv) flameeyes@Gamestation:~/repos/unpaper$ pwd
/home/flameeyes/repos/unpaper
(venv) flameeyes@Gamestation:~/repos/unpaper$ sl status
M imageprocess.c
M imageprocess.h
M meson.build
M unpaper.c
? imageprocess/deskew.c
? imageprocess/deskew.h
```

opening a different folder and going back to the original appears to fix it, but I have no idea why is that even the case.

Flameeyes:
Don't force en_US.utf-8 as locale as it mgiht not exist.

glibc does not generate all combinations of locales by default, and Python
fails when requesting to enforce a locale that does not exist.

Since the extension is not caring for the translation chosen, but rather
only trying to enforce UTF-8 support, use the C.UTF-8 locale which is
guaranteed to exist as per https://www.gnu.org/software/libc/manual/html_node/Choosing-Locale.html

Fixes #814 where VSCode would intermittently be unable to identify
a valid Sapling checkout and/or pull from a repository.

### DIFF
--- a/addons/isl-server/src/Repository.ts
+++ b/addons/isl-server/src/Repository.ts
@@ -1208,7 +1208,7 @@ function getExecParams(
     env: {
       ...options_?.env,
       ...env,
-      LANG: 'en_US.utf-8', // make sure to use unicode if user hasn't set LANG themselves
+      LANG: 'C.UTF-8', // make sure to use unicode if user hasn't set LANG themselves
       // TODO: remove when SL_ENCODING is used everywhere
       HGENCODING: 'UTF-8',
       SL_ENCODING: 'UTF-8',


sggutier:
Thanks for the PR, but we decided to go in another way with a3608fd5cfb5dbe40da372a415e339a92ac724c8 . Sorry for the late notice

PR #877, opened 2024-04-16
ISL: Show tag for PR "follower" commits

rejc2:
ISL: Show a tag for PR "follower" commits (i.e. after running `sl pr follow -r .` on a commit).

image

Currently the `sl` smartlog shows this, but not the interactive smartlog:

image


evangrayk:
Thanks for the PR!

We've been talking about improving the GitHub PR experience. I think making it easier to have multiple commits per PR would be nice, since that's how GitHub works best. I imagine we'll improve this part of the flow as part of that. In the mean time, I think it makes sense to add this given the CLI shows the follower status already.

I think the "Follower" tag works fine on the CLI, but it's maybe a little clunky in the full UI. We can improve that in the future though.

PR #886, opened 2024-04-24
Allow websockets to connect to the current browser's protocol of choice (if available)

randallb:
Allow websockets to connect to the current browser's protocol of choice (if available)

Summary:
Adds a check for the current window's protocol. Defaults to ws:// but will switch to wss:// in secure contexts. This assumes the upstream server will handle converting from wss to ws.

Test plan:
Import github project to replit. Run a server by doing:
`sl web --no-app`

open the replit webview with the token, see if it loads the smartlog.


randallb:
I know the test plan is super weak sauce, but i was having trouble figuring out if it's worth testing against the window object at all...

PR #921, opened 2024-07-17
feat(scm-prompt): add support for dotgit

vegerot:


quark-zju:
FYI There are some internal complaints about the git branch name becomes a commit hash. We run ISL for internal users so `.git/sl` is auto created (by ISL). To address the complaint, I [made](https://github.com/facebook/sapling/commit/9e1993218209dc7fd370e9299c647c2d3082859d) it default to git prompt and can be switched using the `SCM_PROMPT_PREFER_GIT` environment variable.

Issue #906, opened 2024-06-19
ISL: gh: Field 'mergeQueueEntry' doesn't exist on type 'PullRequest'


xEtherealx:
I'm seeing the error gh: Field 'mergeQueueEntry' doesn't exist on type 'PullRequest' when viewing ISL.

I think this is because my enterprise GH version (3.10) doesn't support this field, as it looks to be introduced in 3.12. I saw similar issues filed to the GH pull requests extension, and they worked around it somehow -- I'm guessing by using different/legacy query fields.

```
Failed to fetch Diffs
Command failed with exit code 1: gh api graphql -f searchQuery=repo:... is:pr author:@me -F numToFetch=50 --hostname github.robot.car -f query= query YourPullRequestsQuery($searchQuery: String!, $numToFetch: Int!) { search(query: $searchQuery, type: ISSUE, first: $numToFetch) { nodes { ... on PullRequest { __typename number title body state isDraft url reviewDecision comments { totalCount } mergeQueueEntry { estimatedTimeToMerge } commits(last: 1) { nodes { commit { statusCheckRollup { state } } } } } } } } gh: Field 'mergeQueueEntry' doesn't exist on type 'PullRequest' {"errors":[{"path":["query YourPullRequestsQuery","search","nodes","... on PullRequest","mergeQueueEntry"],"extensions":{"code":"undefinedField","typeName":"PullRequest","fieldName":"mergeQueueEntry"},"locations":[{"line":17,"column":9}],"message":"Field 'mergeQueueEntry' doesn't exist on type 'PullRequest'"}]}
Error: Command failed with exit code 1: gh api graphql -f searchQuery=repo:... is:pr author:@me -F numToFetch=50 --hostname github.robot.car -f query=
    query YourPullRequestsQuery($searchQuery: String!, $numToFetch: Int!) {
  search(query: $searchQuery, type: ISSUE, first: $numToFetch) {
    nodes {
      ... on PullRequest {
        __typename
        number
        title
        body
        state
        isDraft
        url
        reviewDecision
        comments {
          totalCount
        }
        mergeQueueEntry {
          estimatedTimeToMerge
        }
        commits(last: 1) {
          nodes {
            commit {
              statusCheckRollup {
                state
              }
            }
          }
        }
      }
    }
  }
}
    
gh: Field 'mergeQueueEntry' doesn't exist on type 'PullRequest'
[...]

Issue #830, opened 2024-01-30
VS Code Extensions Errors On Getting MergeQueue


zengk95:
Hey all, we're on an older version of Github Enterprise that doesn't have merge queue yet. When using the VS Code Extension, it's trying to fetch the mergequeue fields but it's erroring out. 

![image](https://github.com/facebook/sapling/assets/34172846/e55809a9-8b4b-4126-9f13-99b0eea35e79)

Not sure if it's possible to just fetch it and have it be null or something. We're on a relatively new version (3.9) which is only about 6 months old.

zengk95:
This isn't a blocker per se since I can still use `sl isl` but it's kind of annoying.

alex-statsig:
This is from https://github.com/facebook/sapling/pull/718; you can downgrade the extension in the meantime to unblock too (just released in the latest version). 

To fix I think it would have to retry the query without the field, or detect the GH version and avoid querying it (I think its not that it expects it to be non-null, its that your Github doesnt have that type in its graphql schema at all)

zengk95:
Yea, hopefully it comes out soon. I had to downgrade to 0.1.34 for anyone else having this issue.

radsec:
++...started seeing this a few weeks ago.

zengk95:
Think this is fixed now. Thanks @alex-statsig

PR #926, opened 2024-07-25
Fix handling of GHE without merge queue support

alex-statsig:
Fix handling of GHE without merge queue support

Summary:
https://github.com/facebook/sapling/issues/830 and https://github.com/facebook/sapling/issues/906 reported sapling not working for old Github Enterprise versions since their graphql schema doesnt include merge queue.

This adds a query to detect if merge queue is supported in the graphql version. Then, it will issue a different "YourPullRequestsQuery" depending on if merge queue is supported.

I considered a few other approaches:
- Just detect the error message and retry without merge queue field - a little gross to extract from the error, felt much nicer to explicitly detect the support once
- Use "@include" instead of an entirely separate copy+pasted query - this didn't seem to count as a valid query still via graphiql at least (unsure if the query client here would strip out the unused field though)

Test Plan:
Haven't had a chance to test yet, will try and validate that both versions work and that the detection logic works, but wanted to put up the PR first at least so someone could potentially take over if needed

### DIFF
--- a/addons/isl-server/src/github/generated/graphql.ts
+++ b/addons/isl-server/src/github/generated/graphql.ts
@@ -29334,6 +29334,118 @@ export enum WorkflowState {
   DisabledManually = 'DISABLED_MANUALLY'
 }
 
+/** One possible value for a given Enum. Enum values are unique values, not a placeholder for a string or numeric value. However an Enum value is returned in a JSON response as a string. */
+export type __EnumValue = {
+  __typename?: '__EnumValue';
+  name: Scalars['String'];
+  description?: Maybe<Scalars['String']>;
+  isDeprecated: Scalars['Boolean'];
+  deprecationReason?: Maybe<Scalars['String']>;
+};
+
+/** Object and Interface types are described by a list of Fields, each of which has a name, potentially a list of arguments, and a return type. */
+export type __Field = {
+  __typename?: '__Field';
+  name: Scalars['String'];
+  description?: Maybe<Scalars['String']>;
+  args: Array<__InputValue>;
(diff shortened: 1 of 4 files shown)


luyiyang16:
@muirdm @zzl0 @kavehahmadi60
Could any of you take a look and help merging it if possible. Thanks!

evangrayk:
This looks good to me and works in my testing, but I don't have a github enterprise repo without merge support. Would love help validating that flow works as expected.

evangrayk:
This should be available now in version 0.1.54, please let me know if this works for folks using GHE!

luyiyang16:
I am not using VS Code extension. Which sapling version is this change now available?

luyiyang16:
> This should be available now in sapling-scm VS Code extension version 0.1.54, please let me know if this works for folks using GHE!

I am not using VS Code extension but using isl. Is there a way to patch this to ISL if I don't want to run ISL locally

evangrayk:
> I am not using VS Code extension but using isl. Is there a way to patch this to ISL if I don't want to run ISL locally

I think you either need to build from source or wait for the next Sapling release, which is hopefully coming soon

luyiyang16:
> > I am not using VS Code extension but using isl. Is there a way to patch this to ISL if I don't want to run ISL locally
> 
> I think you either need to build from source or wait for the next Sapling release, which is hopefully coming soon

I tried it locally build from source. Although the error is gone, I am still seeing trouble integrating with GHE - a lot of spinners. Not sure if it is because how I patched it locally doesn't work.

Do you know who should reach out for the next release?

PR #969, opened 2024-10-22
fix(dotgit): set publicheads correctly for transparent git mode

vegerot:
Summary:
0ce7010b7f taught Sapling to use the correct public head based on the upstream
default branch.  However, this doesn't work when using the transparent git mode
through `git clone`.

This diff will correctly set the publicheads config when initializing sapling

Test Plan: see test-git-clone-sets-publicheads.t

### DIFF
--- a/eden/scm/lib/gitcompat/src/config.rs
+++ b/eden/scm/lib/gitcompat/src/config.rs
@@ -22,12 +22,17 @@ impl GlobalGit {
                 &["--show-scope", "--get-regexp", "^(remote|user)\\."],
             )
             .output()?;
-        let out = String::from_utf8(out.stdout)?;
-        Ok(translate_git_config_output(&out))
+        let config = String::from_utf8(out.stdout)?;
+        let remotes_out = self
+            .git_cmd(
+                "ls-remote", &["--symref", ".", "HEAD"])
+                .output()?;
+        let remotes = String::from_utf8(remotes_out.stdout)?;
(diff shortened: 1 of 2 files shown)


eden/scm/lib/gitcompat/src/config.rs
```diff
@@ -22,12 +22,17 @@ impl GlobalGit {
                 &["--show-scope", "--get-regexp", "^(remote|user)\\."],
             )
             .output()?;
-        let out = String::from_utf8(out.stdout)?;
-        Ok(translate_git_config_output(&out))
+        let config = String::from_utf8(out.stdout)?;
+        let remotes_out = self
+            .git_cmd(
+                "ls-remote", &["--symref", ".", "HEAD"])
```
strager, line 28: This performs a network call, which might be unexpected or undesirable.

How about parsing the output of `git rev-parse --symbolic refs/remotes/origin/HEAD` instead?

vegerot: Good idea!  If it's as robust as `ls-remote` then this will save a network call.

quark-zju: You can use [`lookup_reference`](https://github.com/facebook/sapling/blob/a244654809aef840ef06c8dd4d08acf2ce1ac6ba/eden/scm/lib/gitcompat/src/refs.rs#L56) to check the reference without shelling out.

PR #981, opened 2024-11-09
Add --email option to export command for email-style patch formatting

### LINKED ISSUE #318: Patch export
It would be nice if we had the ability to export commits under the Git email .patch (example: https://patch-diff.githubusercontent.com/raw/facebook/sapling/pull/287.patch)

I use the mailing list patch workflow for some personal projects and would love to use sl to manage the repo, but exporting diffs to mail is a bit difficult right now.

`sl export` exists, but it's descended from hg export and does not have all the information the Git patch format does (From lines, date, Subject lines, diffstat, unified file diffs)

VishalGawade1:
**Summary of Changes:**
Fixes #318

**New Flag:** Adds the --email flag to format output with email headers.
**Email Headers:** Generates headers like From, Date, Subject, and Message-Id for each patch to create an email-style format.
**Commit Message and Diff:** Includes the full commit message, followed by diffstat and the full diff for each revision.
**Default Behavior:** If --email is not specified, the command falls back to its standard export functionality without email formatting.


williewillus:
Are you sure you didn't commit stuff by mistake? There's 1862 changed files but the functional change is only in one of them.

VishalGawade1:
> Are you sure you didn't commit stuff by mistake? There's 1862 changed files but the functional change is only in one of them.

Ahh, Apologies for the oversight; I'm not sure how the `/venv` files ended up getting committed.

quark-zju:
There was a Mercurial "patchbomb" extension: https://wiki.mercurial-scm.org/PatchbombExtension

We deleted it since it has no internal use-cases. Would it satisfy your need? If so, maybe we can re-add it so we don't diverge from Mercurial in this use-case.

VishalGawade1:
Really sorry for being little inactive. Re-adding the 'patchbomb' extension sounds like a good to me. It would provide the email functionality in a way that's familiar to Mercurial users.
I can start looking into backporting the 'patchbomb' extension and see how it can be adapted for Sapling. It might take me a while though but any recommendations on how to proceed would be grateful.

PR #1008, opened 2025-01-14
streampager: upgrade dependencies

martinvonz:
Many of streampager's dependencies were very old. The `dirs` version, for example was 5 years old. It seems it's about time to to upgrade.

### DIFF
--- a/eden/scm/lib/third-party/streampager/Cargo.toml
+++ b/eden/scm/lib/third-party/streampager/Cargo.toml
@@ -14,26 +14,26 @@ categories = ["command-line-utilities", "text-processing"]
 name = "streampager"
 
 [dependencies]
-bit-set = "0.5"
-dirs = "2.0"
+bit-set = "0.8.0"
+dirs = "6.0.0"
 enum_dispatch = "0.3.13"
-indexmap = "2.2.6"
-lazy_static = "1.4"
-lru = "0.12.3"
-memmap2 = "0.5.10"
-notify = { version = "5", optional = true }
-regex = "1.9.2"
+indexmap = "2.7.0"
+lazy_static = "1.5.0"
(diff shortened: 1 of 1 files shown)


martinvonz:
I'm trying to go through the internal process to get approval to sign the CLA.

martinvonz:
I've finally gotten the CLA signed, so this is ready for review now. Thanks

quark-zju:
Sorry for taking long. Running tests in the monorepo takes a long time. Some dependencies cannot be upgraded right now:

- `unicode-width`: blocked by https://github.com/phsym/prettytable-rs/issues/165. Ideally `prettytable-rs` should pin `unicode-width` itself so the global `unicode-width` can be upgraded.
- `bit-set`: upgrade caused hhvm to fail compile.
- `memmap2`: upgrade caused a hhvm test to segfault ("address not mapped to object").
- `notify`: 8.0 does not compile on macos - it uses `fsevents_sys::FSEventsGetCurrentEventId` introduced by fsevents-sys 4.1.0 but notify's Cargo.toml specifies `fsevents-sys = "4.0.0"`.
- `lru`, `indexmap`: upgrade introduces `hashbrown 0.15.2` which does not compile with "nightly" feature turned on, because it uses the `strict_provenance_lints` feature introduced by Rust 1.84 (Jan 8). We can retry after the internal Rust team upgrade the compiler.

I'll upgrade the rest once tests pass and publish a new release.

ilyagr:
Is `termux` difficult to update? TODO: Link to the relevant thread about NetBSD.

quark-zju:
`termwiz 0.23` upgrade is not in this change but I can try to get it included.

quark-zju:
I published a new version of streampager that includes the `termwiz` upgrade.

PR #1017, opened 2025-01-22
docs: missing code block in pull.md

JamBalaya56562:
## Summary

Add missing code block to [pull docs](https://sapling-scm.com/docs/commands/pull/).

### DIFF
--- a/eden/scm/sapling/commands/__init__.py
+++ b/eden/scm/sapling/commands/__init__.py
@@ -4752,7 +4752,7 @@ def pull(ui, repo, source="default", **opts):
 
           @prog@ pull
 
-      - pull a bookmark named my-branch from source my-fork:
+      - pull a bookmark named my-branch from source my-fork::
 
           @prog@ pull my-fork --bookmark my-branch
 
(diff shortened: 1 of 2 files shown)


sggutier:
Thanks for the fix @JamBalaya56562 , but this doc comes from this comment actually: https://github.com/facebook/sapling/blob/main/eden/scm/sapling/commands/__init__.py#L4746

The documentation for the commands in the website actually is generated through the documentation for the CLI commands. Could you modify the command instead, and then regenerate the entire website as appropriate?

JamBalaya56562:
> Thanks for the fix @JamBalaya56562 , but this doc comes from this comment actually: https://github.com/facebook/sapling/blob/main/eden/scm/sapling/commands/__init__.py#L4746
> 
> The documentation for the commands in the website actually is generated through the documentation for the CLI commands. Could you modify the command instead, and then regenerate the entire website as appropriate?

Thank you for waiting very much!
I have regenerated the [pull.md](https://github.com/facebook/sapling/blob/main/website/docs/commands/pull.md) finally.

PR #1035, opened 2025-02-28
Chore: address various misspeelings

vincent-4:
Just for fun, I went ahead and ran codespell, with manual review of course. It's pretty inaccurate in edenFS. The original goal was to make docs / site / sl commands not have typos.

This non-breaking where possible. Tbh, at an average rate of ~0.18 typos a day on sapling, very good!

This might? shock everyone working on a modified file but it should? be fine.

Using `codespell --skip "*.po,*.ts,./src/3rdParty,./src/Test,./lib" --count --quiet-level 3 -w --ignore-words-list "vertices,vertexes,tolen,anc,re-use,pres,preformed,ser,Abomonation,abomonation"`

Last question: would FB invite having codespell as a pre-commit hook? Could be useful, even only as warnings.


vincent-4:
I'm going to infer this is expected 
> Verify addons/ folder: All jobs have failed

quark-zju:
Thanks. We do have a spell check linter internally but it is obviously not as powerful...

PR #1067, opened 2025-04-26
make `single` pr workflow default for `sl pr submit`

discentem:


eden/scm/sapling/ext/github/submit.py
```diff
@@ -72,11 +72,11 @@ def from_config(ui) -> "SubmitWorkflow":
         workflow = ui.config(
             "github", "pr-workflow", ui.config("github", "pr_workflow")
         )
-        if not workflow or workflow == "overlap":
+        if not workflow or workflow == "single":
             # For now, default to OVERLAP.
```
parag-gangil-cko, line 76: The default is no longer 'OVERLAP'
```suggestion
            # For now, default to SINGLE.
```

PR #1101, opened 2025-07-02
OSS edenfs_linux build fixes

ben--:
This change should be landed after or instead of #1100 (it includes that as the first commit)

This PR fixes a number of build issues in the OSS build. Each individual fix can be verified by running:

```
act --container-architecture linux/x86_64 -W ".github/workflows/edenfs_linux.yml"
```

on the prior commit. (Though, due to job prioritization, sometimes you may find a different issue first.)

The above build still does not complete due to a lack of thrift file generation, but that modification will be largely independent of this "stack".

### DIFF
--- a/CMake/EdenConfigChecks.cmake
+++ b/CMake/EdenConfigChecks.cmake
@@ -110,6 +110,7 @@ endif()
 
 find_package(python-toml REQUIRED)
 find_package(python-filelock REQUIRED)
+find_package(python-psutil REQUIRED)
 
 # pexpect is used by some of the integration tests.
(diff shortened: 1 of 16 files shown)


quark-zju:
@ben-- Could you help rebasing this change? I made some changes to satisfy the linters and got some conflicts rebasing this diff - It'd be easier if ShipIt can import a stack as a stack instead of one commit... unfortunately we don't have the right tooling right now. We use `black` for Python format and `clangformat` for C++ format.

ben--:
Made unnecessary by #1125-1134

PR #1105, opened 2025-07-11
fix: Add missing thrift "bareclient" dependency

ben--:


quark-zju:
This `bareclient` actually seems quite complicated - bringing in tens of dependencies... So it's unlikely fixable by just exposing this directory.

PR #1132, opened 2025-09-23
fix(oss): Resolve a number of build issues only visible in OSS

ben--:
This PR contains a number of small build fixes -- mostly areas where the OSS EdenFS build had not kept up to date with the internal implementation.

Tested through building Sapling/EdenFS (requires other fixes to run clean):
```
act --container-architecture linux/x86_64 -W .github/workflows/edenfs_linux.yml
```


bherila:
Thank you for this, I will review it as soon as I can

PR #1200, opened 2026-02-04
feat(isl): PR Review Mode - foundation for phases 9-14

sontiO:
This commit contains the foundational work for the ISL PR Review Mode feature:
- Phase 9: Review mode state and entry points
- Phase 10: Inline comments and threading
- Phase 11: Review submission
- Phase 12: Merge + CI status
- Phase 13: Sync/rebase support
- Phase 14: Stacked PR navigation

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/1200).
* #1204
* #1203
* #1202
* #1201
* __->__ #1200

PR #1201, opened 2026-02-04
fix(isl): prevent white flash on load and remove debug artifacts

sontiO:
- Add dark background color to html/body to prevent white flash on load
- Remove LOCAL DEV BUILD debug banner from App.tsx

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/1201).
* #1204
* #1203
* #1202
* __->__ #1201
* #1200


sontiO:
Created on wrong repo

PR #1208, opened 2026-02-05
[RFC] Fix reviewstack.dev deploy by removing broken Docker container dependency

ajaymenon:
# Summary

The repo used to have a CI workflow that built an Ubuntu 22.04 Docker image (generated by `ci/gen_workflows.py` as `sapling-cli-ubuntu-22.04-image.yml`) and pushed it to GitHub Container Registry as `ghcr.io/facebook/sapling/build_ubuntu_22_04:latest`. At some point, that workflow and its Dockerfile were removed from the repo, but two workflows that depend on that image were never updated:

- `reviewstack.dev-deploy.yml`
- `sapling-addons.yml`

So every time either of those workflows runs, it tries to pull an image that no longer exists, and fails immediately at container initialization.

Example failure on `main`: https://github.com/facebook/sapling/actions/runs/21693937235/job/62559938248

# Test Plan
Not 100% sure how to test this tbh.  I tried to add a trigger to get GHA to pick up the change and see if the deploy action would succeed but it looks like I don't have permissions to do that.

ubuntu-22.04 should already have node and yarn so the build deployment should work without any custom dependencies.

### DIFF
--- a/.github/workflows/reviewstack.dev-deploy.yml
+++ b/.github/workflows/reviewstack.dev-deploy.yml
@@ -2,26 +2,29 @@ name: Publish https://reviewstack.dev
 
 on:
   workflow_dispatch:
+  pull_request:
+      paths:
+        - '.github/workflows/reviewstack.dev-deploy.yml'
   schedule:
     - cron: '0 0 * * 1-5'
 
 jobs:
   deploy:
     runs-on: ubuntu-22.04
-    # Our build container already has Node, Yarn, and Python installed.
-    container:
-      image: ${{ format('ghcr.io/{0}/build_ubuntu_22_04:latest', github.repository) }}
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v6
-      - name: Grant Access
-        run: git config --global --add safe.directory "$PWD"
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
+      - name: Install Yarn
+        run: npm install -g yarn
       - name: Install dependencies
         working-directory: ./eden/contrib/
-        run: yarn install --prefer-offline
-
+        run: yarn install
       # Build codegen and then do some sanity checks so we don't push the site
       # when the tests are broken.
       - name: GraphQL/TextMate codegen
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewstack.dev-deploy.yml
```diff
@@ -1,29 +1,30 @@
 name: Publish https://reviewstack.dev
 
-on:
+on:
```
bherila: Please remove trailing whitespace

bherila (changes requested):
Could I request you restore the comments and remove trailing whitespace, I will check tomorrow to see how we can validate the diff. I think it is OK however, just these nits :)

bherila:
Btw thank you for your contribution to reviewstack!

ajaymenon:
Also still getting used to Github as opposed to Phabricator - I've made the requested changes and was able to re-request review, but it still says the PR has changes requested.

bherila:
Thanks, I will get to it today anyway!

ajaymenon:
@bherila - it seems that Merging is blocked because "Merge is not an allowed merge method in this repository". Not quite sure how else I'm supposed to commit this PR. Would you be willing to take it over?

The reason the local build+deploy method doesn't solve my problem is because my colleagues are the ones trying to use reviewstack to review my PRs internally.

bherila:
Because I made a change to this diff internally, I need an additional review from another employee. If any Meta employee sees this and can merge the diff, please do! I will ping some more people today if not.

bherila:
By the way @ajaymenon 

1. if you make a new PR for this there is a good chance I can merge it without my changes, then I can rebase this one (sorry, even if I undo my changes on this one I will still need another employee's review)
2. check out #1213 , #1214 which might be waiting for anyway :)

but I'm asking around! (fyi, Reviewstack is a "extracurricular" project, It's not my primary job.)

ajaymenon:
I would love to make a new PR right now but the github outages are making that impossible on my side. Also no pressure I get it's a side thing - appreciate your followups here!

@ashwinb maybe you could help approve this PR from the other side?

ajaymenon:
Nvm was able to make https://github.com/facebook/sapling/pull/1216

bherila:
OK I merged your PR. I am going to close this. 

However, I will follow up tomorrow with the eslint fixes that should unblock the deploy

Thank you @ajaymenon !

bherila:
Merged the other one, closing this

PR #1210, opened 2026-02-06
[RFC] Expose --pr-workflow CLI option to switch between overlap (default) and single (stacked) behavior for github PR submissions

ajaymenon:
# Summary
The default `sl pr submit` creates overlapping commits where each PR on github shows all the changes including parents instead of just the deltas. reviewstack.dev makes this process easier but it also sometimes falls over when I have a stack of 15+ commits. I realized that the config option `sl -config github.pr-workflow=single pr submit` exists and creates one PR for each commit where github recognizes the base branch for each PR as being its parent commit in sapling. 

I wanted to expose this option in the CLI a bit more easily, but I'm not sure of the best way to do it. Please advise.

# Test Plan
New unit tests:
`cd eden/scm/tests && python3 run-tests.py test-ext-github-pr-submit-pr-workflow.t`

Also tried this in a separate repo where I did
```
touch 1.txt
sl commit -m "1.txt"
touch 2.txt
sl commit -m "2.txt"
 sl --config github.pr-workflow=single pr submit
```

and github created two PRs, where 1 was on top of main and 2 was on top of 1.

Ruff found some existing formatting issues in some of the modified files - I'm not sure if this is because my environment is not configured properly though.

### DIFF
--- a/eden/scm/sapling/ext/github/__init__.py
+++ b/eden/scm/sapling/ext/github/__init__.py
@@ -120,21 +120,6 @@ def submit_cmd(ui, repo, *args, **opts):
 
     Returns 0 on success.
     """
-    pr_workflow = opts.get("pr_workflow")
-    if pr_workflow:
-        normalized_pr_workflow = pr_workflow.strip().lower()
-        if normalized_pr_workflow not in {"overlap", "single"}:
-            raise error.Abort(
-                _(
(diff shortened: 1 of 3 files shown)


ajaymenon:
ah nvm i can just set this option in my own config instead.

PR #1278, opened 2026-04-09
sl push: more helpful message

stepancheg:
```
sl push   # no args
```

```
abort: nothing to push - current commit is already at origin/master
```

instead of

```
abort: use '--to' to specify destination bookmark
```


quark-zju (changes requested):

quark-zju:
(seems comment isn't synced so I pasted it here)

The code seems fragile - it currently depends on the subtle behavior that `_guesspushtobookmark` returns None for "nothing to push" case, but `_guesspushtobookmark` could also return None for other reasons. It also prints all remotenames of the current commit, instead of the intended implicit "to".

How do you think about?
1. Change `_guesspushtobookmark` report the "origin/foo" even if it is the current commit.
2. (Optionally) Names to print (`at_bookmarks`) use the explicit or implicit "to", not listing what names the current commit has.
3. (Optionally) If nothing pushed (note the local remotenames can be outdated, so attempt to push is still needed), exit 1 instead of 255.

Issue #787, opened 2023-11-27
Can't update github workflows (refusing to allow an OAuth App to create or update workflow without `workflow` scope)


terminalnode:
```sh
sl pr submit --draft
pushing 1 to https://github.com/MyOrg/MyProject
abort: `git --git-dir /path/to/project/.sl/store/git push --force https://github.com/MyOrg/MyProject a443cbe2450bd5ea5ae1ae080c75d6cd128675d3:refs/heads/pr596` failed with exit code 1: stdout: 
stderr: To https://github.com/MyOrg/MyProject
 ! [remote rejected]     a443cbe2450bd5ea5ae1ae080c75d6cd128675d3 -> pr596 (refusing to allow an OAuth App to create or update workflow `.github/workflows/linter.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/MyOrg/MyProject'
```
This happened when I amended a commit to make changes to our workflows. Is there any way around it?

PR #1299, opened 2026-05-08
docs: document workflow OAuth scope requirement for sl pr submit

Marukome0743:
The default scopes requested by `gh auth login` (`repo`, `read:org`,
`gist`) do not include `workflow`. As a result, `sl pr submit` aborts
with an `UNAUTHORIZED-WRITE` error from the GraphQL `mergeBranch`
mutation whenever a commit touches `.github/workflows/`. Because the
SSH push and PR creation succeed first, the PR exists by the time the
command exits non-zero, which makes the failure mode hard to debug.

Add a `:::note` near the auth instructions pointing at
`gh auth refresh -s workflow`, and a Troubleshooting subsection that
walks through the full error and the order of operations. Cross-link
issue #787, which is the same root cause surfaced via HTTPS push.


quark-zju (approved):

PR #1336, opened 2026-06-12
[RFC] Add GitHub PR update delta comments

ajaymenon:
## Context
I really like using sapling, but the thing I can't stand about it when using github is that when I amend a commit, the github UI doesn't treat it as a delta (but the reviewstack UI does). This is a problem when I am responding to feedback on a PR where the reviewer prefers to use the Github UI - the reviewer can't easily view the delta between the revision they commented on and the updated revision on the PR after I address their feedback (Github sees it as one single commit even though on the backend sapling is tracking the commit hashes).

## Proposed Solution
Track the SHAs of each PR head, and post comments that show the delta between the commits.

1. Sapling builds the submit plan like usual.
2. Opt-in `github.pr-submit-comment-deltas` flag. When enabled, compare each existing PR's previous head SHA with the new commit SHA. 
3. For each changed PR, build a comment containing the github explore comparison link.
4. Sapling archives the old head into the existing `sapling-pr-archive-*` branch so that the commits are still resolvable by github. (Open question - are there downstream implications of this that I don't quite understand?)
5. Sapling does the rest of the normal push flow for the stack.
6. After the submit succeeds, we post the delta comment(s).

## Open Questions
1. I went with the "comment" approach since it seemed less risky than rewriting PR body or existing user comments - not sure if there's a better approach. Is a comment the right thing to do?
2. How about for stacks? Should we post one comment per PR or one summary comment for the stack? I could see github ratelimiting becoming a problem or an issue (also why currently I serially post the comments and wait 1s in between).
3. What should we do if the archiving of the old head fails? Currently I warn and continue, but that means the compare link might not be useful.

## Tests
1. Unit tests for
  a. Make two commit stack with PRs, amend top comment, resubmit stack. Assert only top PR gets a comment.
  b. Validate that we are archiving old PR heads properly, commenting after the archive is done, etc.
2. Integration tests - I built the sapling binary locally and used it to make this PR. You can see the sapling posted comments with deltas below :D

### DIFF
--- a/eden/scm/sapling/ext/github/consts/query.py
+++ b/eden/scm/sapling/ext/github/consts/query.py
@@ -81,6 +81,18 @@ mutation ($pullRequestId: ID!, $title: String!, $body: String!, $base: String!) 
 }
 """
 
+GRAPHQL_ADD_COMMENT = """
+mutation ($subjectId: ID!, $body: String!) {
+  addComment(input: {subjectId: $subjectId, body: $body}) {
+    commentEdge {
+      node {
+        id
+      }
+    }
+  }
+}
+"""
+
 GRAPHQL_CREATE_BRANCH = """
 mutation ($repositoryId: ID!, $name: String!, $oid: GitObjectID!) {
   createRef(input: {repositoryId: $repositoryId, name: $name, oid: $oid}) {
--- a/eden/scm/sapling/ext/github/gh_submit.py
+++ b/eden/scm/sapling/ext/github/gh_submit.py
@@ -355,6 +355,20 @@ async def update_pull_request(
         return Ok(result.unwrap()["data"]["updatePullRequest"]["pullRequest"]["id"])
 
 
+async def add_comment(hostname: str, subject_id: str, body: str) -> Result[str, str]:
+    """Adds a comment to an issue or pull request, returning the comment node ID."""
+    params: Dict[str, _Params] = {
+        "query": query.GRAPHQL_ADD_COMMENT,
+        "subjectId": subject_id,
+        "body": body,
+    }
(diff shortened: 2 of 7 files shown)


ajaymenon:
[//]: # (sapling-pr-delta old=ceee4af613bddc794318a8a9e8ae4f839e7a7abf new=49269af1b4588e01ccd8475b503b70ecf3158589)

Sapling updated this PR.

Review delta: https://github.com/facebook/sapling/compare/ceee4af613bddc794318a8a9e8ae4f839e7a7abf..49269af1b4588e01ccd8475b503b70ecf3158589
Previous head: `ceee4af613bd`
New head: `49269af1b458`

ajaymenon:
[//]: # (sapling-pr-delta old=49269af1b4588e01ccd8475b503b70ecf3158589 new=d01871c6c4a28ccc090deccd3ffff342118c39b4)

Sapling updated this PR.

Review delta: https://github.com/facebook/sapling/compare/49269af1b4588e01ccd8475b503b70ecf3158589..d01871c6c4a28ccc090deccd3ffff342118c39b4
Previous head: `49269af1b458`
New head: `d01871c6c4a2`

ajaymenon:
@bherila hello again! This is something that my colleagues have mentioned they find useful (a few of us are using sapling internally, but nobody is really using reviewstack UI yet due to performance and nobody wanting to go through the trouble of self-hosting it).

ajaymenon:
When adding the initial logic to submit.py, linter complained due to the function's complexity being pretty high so I split it up into separate functions. https://app.semanticdiff.com/gh/facebook/sapling/pull/1336/changes#eden/scm/sapling/ext/github/submit.py?ignore_comments=false makes it a bit clearer that it's largely a move-refactoring.

PR #1364, opened 2026-07-21
streampager + curses_compat: fix Shift keys (G, @, capitals, A) broken in Ghostty

### LINKED ISSUE #1363: 🐛bug: builtin pager ignores Shift keys in Ghostty
## Description

When using Sapling's internal pager (`pager=internal:streampager`, the default for `sl log`), any key that requires Shift does not work when running **outside of tmux in Ghostty.app** (and cmux.app which uses libghostty).

Observed failures:
- `G` (Shift+g) to go to bottom of pager does nothing
- Typing `@` (Shift+2 on US layout) in search prompt (`/`) results in empty input
- Searching for any capital letter (e.g. `Foo`) does not insert the capital
- All shifted punctuation / symbols affected

Inside tmux it works (tmux re-encodes keys as plain chars), and inside Terminal.app it works. Only Ghostty (and libghostty-based terms) outside tmux reproduces.

## Repro steps (Ghostty, no tmux)

1. `cd ~/code/github.com/facebook/sapling && sl log -l 5000 --pager=always`
2. Press lowercase `g` -> goes to top (works)
3. Press uppercase `G` (Shift+g) -> **should go to bottom, stays at top**
4. Press `/` to search
5. Type `@` (Shift+2) -> **prompt stays empty, should show @**
6. Type `ABC` -> **nothing inserted or only lowercase**

## Repro with hex injection (works inside tmux too, to avoid needing Ghostty)

Ghostty with `modifyOtherKeys=2` enabled (which Sapling enables via termwiz `ESC[>4;2m` in `set_raw_mode`) encodes shifted keys as xterm modifyOtherKeys:

- Shift+G: `ESC[27;2;71~`  hex `1b 5b 32 37 3b 32 3b 37 31 7e`
- Shift+@: `ESC[27;2;64~`  hex `1b 5b 32 37 3b 32 3b 36 34 7e`

```bash
tmux new -d -s repro -x 200 -y 20
tmux send-keys -t repro "sl log -l 5000 --pager=always" C-m; sleep 2
tmux send-keys -t repro -l "g"; sleep 0.5    # top works
tmux send-keys -t repro -H '1b5b32373b323b37317e'; sleep 0.8  # Shift+G Ghostty
# expected bottom, actual stays at top -> bug
tmux send-keys -t repro -l "/"; sleep 0.3
tmux send-keys -t repro -H '1b5b32373b323b36347e'; sleep 0.6  # @ shift
# expected Search: @, actual empty
tmux send-keys -t repro "q"; tmux kill-session -t repro
```

## Technical overview

Sapling's pager enables xterm modifyOtherKeys and Ghostty sends Shift+G as ESC[27;2;71~ which termwiz decodes as Char('G')+SHIFT. The pager's keymap expects Char('G')+NONE and prompt only accepts NONE, so the exact modifier check fails. tmux hides this by re-encoding to plain G.

**Disclosure**: Written by Meta Muse Spark 1.1

vegerot:
Summary:
When running under Ghostty (and cmux which uses libghostty) outside of tmux,
eden/scm's internal pager and interactive commit UI did not register any
Shift-modified key.

Root cause:
- `termwiz` `UnixTerminal::set_raw_mode()` enables `modifyOtherKeys=2` (`ESC[>4;2m`).
- Ghostty's `key_encode.zig` correctly implements this: `Shift+G` is sent as
  `ESC[27;2;71~` (where 71 is uppercase G) and `Shift+@` as `ESC[27;2;64~`.
- `termwiz` `InputParser` decodes these as `Char('G')+SHIFT` and `Char('@')+SHIFT`.

Two separate code paths were affected:

1. **streampager** (Rust, used by `sl log` etc.):
   The keymap stores `'G'` as `(NONE, Char('G'))`.  The lookup used exact
   modifier match, so `(SHIFT, Char('G'))` never matched — `G` was silently dropped.

2. **curses_compat** (Python, used by `sl commit --interactive` / crecord):
   `curses_compat.py:getch()` only handled `Modifiers.NONE` and `Modifiers.CTRL`.
   When `SHIFT` was set, it fell through to `return -1`, silently dropping the key
   so `A` (toggle all), `G`, `J`, `H`, `F` etc. did not work.
   Additionally, the CTRL check used `Modifiers.CTRL or Modifiers.RIGHT_CTRL` (Python
   `or` always evaluates to `Modifiers.CTRL`); fixed to use bitwise `|` with `&`.

tmux masks this because `send-keys -l` sends plain `0x47` (`G+NONE`). Raw hex
injection via tmux reproduces the bug:
```
tmux send-keys -H '1b5b32373b323b37317e'  # ESC[27;2;71~ Shift+G
```
Old `sl` stays at top, new `sl` goes to bottom.

Fix:
Strip the SHIFT modifier and retry the lookup:

1. **streampager**: In `Screen::dispatch_key` and `Prompt::dispatch_key`, build a
   candidate list that includes the original key plus a SHIFT-stripped variant, then
   try each against the keymap/prompt match.  `modifyOtherKeys=2` already sends the
   shifted character (uppercase), so stripping SHIFT is sufficient.

2. **curses_compat**: In `getch()`, add a `modifiers & SHIFT` branch that returns
   `ord(ch.upper())`. Also fix the CTRL modifier check from `or` to bitwise `|`.

Test Plan:
- `make oss` - builds (but requires #1365)
- tmux test:
  ```bash
  tmux new -d -s t; tmux send-keys -t t './out/sl log -l 5000 --pager=always'
  # Check plain G goes to bottom, g to top.
  tmux send-keys -t t -H '1b5b32373b323b37317e'  # Shift+G Ghostty modifyOtherKeys
  # => now goes to bottom (was stuck at top)
  tmux send-keys -t t -l '/' ; send -H '1b5b32373b323b36347e' # Shift+@ in prompt
  # => now shows Search: @ (was empty)
  ```
  Verified old `/usr/local/bin/sl` fails, new `out/sl` passes.
- `cargo test -p sapling-streampager` still 5 passed.
- `python tests/run-tests.py tests/test-commit-interactive-curses.t` passes.
- Manual in Ghostty.app outside tmux: `G` jumps to bottom, `/` `@` search works,
  capital letters searchable (previously broken). `sl commit --interactive` `A`
  toggles all files (previously broken).

Fixes #1363
Co-authored-by: Opencode <opencode@anomalyco.dev>

### DIFF
--- a/eden/scm/lib/third-party/streampager/src/prompt.rs
+++ b/eden/scm/lib/third-party/streampager/src/prompt.rs
@@ -5,7 +5,7 @@ use std::fmt::Write;
 
 use termwiz::cell::{AttributeChange, CellAttributes};
 use termwiz::color::{AnsiColor, ColorAttribute};
-use termwiz::input::KeyEvent;
+use termwiz::input::{KeyCode, KeyEvent};
 use termwiz::surface::change::Change;
 use termwiz::surface::Position;
 use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
@@ -371,6 +371,13 @@ impl Prompt {
         const CTRL: Modifiers = Modifiers::CTRL;
         const NONE: Modifiers = Modifiers::NONE;
         const ALT: Modifiers = Modifiers::ALT;
+        // Strip SHIFT from Char keys (modifyOtherKeys=2 sends shifted char + SHIFT).
+        let mut key = key;
+        if key.modifiers.contains(Modifiers::SHIFT) {
+            if let KeyCode::Char(_) = key.key {
+                key.modifiers.remove(Modifiers::SHIFT);
+            }
+        }
         let value_width = width - self.prompt.width() - 4;
         let action = match (key.modifiers, key.key) {
             (NONE, Enter) | (CTRL, Char('j')) | (CTRL, Char('m')) => {
--- a/eden/scm/lib/third-party/streampager/src/screen.rs
+++ b/eden/scm/lib/third-party/streampager/src/screen.rs
@@ -1241,17 +1241,38 @@ impl Screen {
         key: KeyEvent,
         event_sender: &EventSender,
     ) -> DisplayAction {
-        if let Some(binding) = self.keymap.get(key.modifiers, key.key) {
+        if let Some(binding) = Self::key_candidates(key)
+            .into_iter()
+            .find_map(|key| self.keymap.get(key.modifiers, key.key))
+        {
             match binding {
                 Binding::Action(action) => {
                     let action = action.clone();
                     return self.dispatch_action(action, event_sender);
                 }
-                Binding::Custom(b) => b.run(self.file.index()),
+                Binding::Custom(b) => {
+                    b.run(self.file.index());
+                    return DisplayAction::Render;
+                }
                 Binding::Unrecognized(_) => {}
             }
         }
         DisplayAction::Render
+    }
+
+    fn key_candidates(key: KeyEvent) -> Vec<KeyEvent> {
+        use termwiz::input::{KeyCode, Modifiers};
+
+        let mut candidates = vec![key.clone()];
+        if matches!(key.key, KeyCode::Char(_)) && key.modifiers.contains(Modifiers::SHIFT) {
+            let mut modifiers = key.modifiers;
+            modifiers.remove(Modifiers::SHIFT);
+            candidates.push(KeyEvent {
+                key: key.key,
+                modifiers,
+            });
+        }
+        candidates
     }
 
     /// Append a digit to the repeat count.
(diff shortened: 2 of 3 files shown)


eden/scm/lib/third-party/streampager/src/prompt.rs
```diff
@@ -371,6 +371,13 @@ impl Prompt {
         const CTRL: Modifiers = Modifiers::CTRL;
         const NONE: Modifiers = Modifiers::NONE;
         const ALT: Modifiers = Modifiers::ALT;
+        // Strip SHIFT from Char keys (modifyOtherKeys=2 sends shifted char + SHIFT).
```
quark-zju, line 374: Is this strip applied too widely (e.g. cannot distinguish between Shift+Tab vs Tab)? Should it be scoped to keys that have a known "shifted" state (e.g. lower-case letters, main row numbers, and some other ANSI keys)

vegerot: Yeah, I was thinking about that too.  It seems like there's a lot of edge cases here, especially when you consider non-standard keyboard layouts.  For example, I use a programmer Dvorak layout where my number row is (unshifted): `+[{(&=)}]*` and _shifted_ is: `1234567890`.

This would be simpler if https://github.com/wezterm/wezterm/issues/7977 were resolved.

For now, I don't think this patch is introducing any new issues.

PR #1365, opened 2026-07-21
fix(oss-build): allow nightly features on stable via RUSTC_BOOTSTRAP

vegerot:
make oss failed on stable Rust 1.97.1 with:

  error[E0554]: #![feature] may not be used on stable
    --> smallvec-1.15.2/src/lib.rs:95: specialization
  and 8 crates using #![feature(once_cell_try)]

Buck builds use a nightly toolchain internally and handle this,
but cargo oss builds use stable rustc by default.

Cargo.toml files are @generated by autocargo from BUCK files,
so we should not hand-edit them to remove the features.

Setting RUSTC_BOOTSTRAP=1 in build.py's cargo_env allows stable
cargo/rustc to accept these nightly features without requiring
users to install a nightly toolchain or override.

Fixes: cd eden/scm && make -j oss

### DIFF
--- a/eden/scm/build.py
+++ b/eden/scm/build.py
@@ -430,6 +430,9 @@ def cargo_env(args):
         env["OPENSSL_DIR"] = str(openssl)
         env.pop("OPENSSL_STATIC", None)
     env["RUSTFLAGS"] = (env.get("RUSTFLAGS", "") + " -Anon_local_definitions").strip()
+    # Allow nightly features (smallvec specialization, once_cell_try) on stable Rust.
+    # Required for `make oss` / cargo oss builds. Buck builds use nightly toolchain internally.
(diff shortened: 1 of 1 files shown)


quark-zju:
According to https://doc.rust-lang.org/beta/unstable-book/compiler-environment-variables/RUSTC_BOOTSTRAP.html

> For library crates, we especially discourage the use of this feature. The crates depending on you do not know that you use this feature, have little recourse if it breaks, and can be used in contexts that are hard to predict.

Part of the codebase is published on crates.io. While this env var solves the Sapling binary build, it will force downstream users of the Sapling libraries to use nightly, which is undesirable.

I've sent diffs internally to avoid the nightly features.

vegerot:
@quark-zju thanks!  I'll close this PR when the other fix lands 🙂

PR #1366, opened 2026-07-22
pager: add opt-in mouse wheel scrolling to streampager

DurhamG:
Summary:
This adds a `pager.mouse` config option, off by default, that turns mouse reporting on and scrolls the output three lines per wheel click.

It is off by default because it interferes with normal terminal selection and copying behavior.

The option is named `pager.mouse` even though it only affects streampager, because that is what the existing options do (`pager.wrapping-mode`, `pager.interface`, etc). The new config is documented as "Only affects streampager".

The display event loop turns a wheel event into the same scroll actions the key bindings already use. That loop checks the config option rather than relying on mouse events not arriving at all, because on Windows termwiz always puts the console into mouse-input mode, so wheel events show up even when the user did not ask us to handle the mouse.

Sapling passes the option in when it builds the pager rather than through a setter, because the value decides which terminal capabilities the pager is created with. streampager also reads the option from an SP_MOUSE environment variable for use outside Sapling, matching the existing SP_INTERFACE_MODE and SP_SCROLL_PAST_EOF.

Tested by hand on Windows and macOS with `sl help config --config pager.mouse=true` and `sl help config --config pager.mouse=false`, confirming the wheel scrolls only when it is on.

### DIFF
--- a/eden/scm/lib/io/src/lib.rs
+++ b/eden/scm/lib/io/src/lib.rs
@@ -31,6 +31,7 @@ use pipe::pipe;
 use spawn_ext::CommandExt;
 use streampager::Pager;
 use streampager::action::Action;
+use streampager::config::Config as PagerConfig;
 use streampager::config::InterfaceMode;
 use streampager::config::WrappingMode;
 use term::DEFAULT_TERM_HEIGHT;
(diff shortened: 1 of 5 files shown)


DurhamG:
AI was used in the making of this commit.

PR #1392, opened 2026-07-31
[1/n][sl][github][gh stacks] support headers and list params

raydatray:
### ctx
github added native stacked pull requests, it would be nice to add native support to it for sapling!

this stack lets `sl pr submit` and `sl pull` via a new `github.pr.workflow = stacked` mode

### changes made
- `github_gh_cli.make_request` accepts custom headers (`-H`, needed for the `X-GitHub-Api-Version` preview header)
- `_format_param` supports list values using the `gh api` repeated-field syntax (`pull_requests[]=101`)
- an empty list is passed explicitly as `key[]` without a value (per the gh api manual) rather than dropped, since an empty array and a missing field can mean different things to an endpoint
- mock_utils: `MockGitHubServer` now tracks which expectations were consumed; tests can opt in via `wrap_with_consumption_check` to fail when an expected request silently stops happening (closes the old TODO)

### test plan
doctests in `_format_param` (registered in test-doctest.py, including the empty-array case), existing github .t suite. verified the empty-array wire format empirically: `gh api -F "pull_requests[]" --verbose` sends `{"pull_requests": []}`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/sapling/pull/1392).
* #1398
* #1397
* #1396
* #1395
* #1394
* #1393
* __->__ #1392

### DIFF
--- a/eden/scm/sapling/ext/github/github_gh_cli.py
+++ b/eden/scm/sapling/ext/github/github_gh_cli.py
@@ -16,32 +16,43 @@ from sapling.result import Err, Ok, Result
 
 JsonDict = Dict[str, Any]
 
+# Scalar value that can be passed as a field to `gh api`.
+_ScalarParam = Union[str, int, bool]
+# `gh api` also supports array fields via repeated `key[]=value` args.
+ParamValue = Union[_ScalarParam, List[_ScalarParam]]
+
 
 async def make_request(
-    params: Dict[str, Union[str, int, bool]],
+    params: Dict[str, ParamValue],
     hostname: str,
     endpoint="graphql",
     method: Optional[str] = None,
+    headers: Optional[Dict[str, str]] = None,
 ) -> Result[JsonDict, str]:
     """If successful, returns a Result whose value is parsed JSON returned by
     the request.
     """
-    return await _make_request(params, hostname, endpoint, method)
+    return await _make_request(params, hostname, endpoint, method, headers)
 
 
 # Unexported extension/mock point.
 async def _make_request(
-    params: Dict[str, Union[str, int, bool]],
+    params: Dict[str, ParamValue],
     hostname: str,
     endpoint: str,
     method: Optional[str],
+    headers: Optional[Dict[str, str]] = None,
 ) -> Result[JsonDict, str]:
     if method:
         endpoint_args = ["-X", method.upper(), endpoint]
     else:
         endpoint_args = [endpoint]
+    header_args = list(
+        itertools.chain(*[["-H", f"{k}: {v}"] for (k, v) in (headers or {}).items()])
+    )
     args = (
         ["gh", "api", "--hostname", hostname]
+        + header_args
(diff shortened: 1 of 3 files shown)


eden/scm/sapling/ext/github/github_gh_cli.py
```diff
@@ -82,7 +93,30 @@ async def _make_request(
...
+    ['-F', 'pull_requests[]=101', '-F', 'pull_requests[]=102']
+    >>> _format_param("labels", ["bug", "help wanted"])
+    ['-f', 'labels[]=bug', '-f', 'labels[]=help wanted']
+    >>> _format_param("empty", [])
+    []
+    """
+    if isinstance(value, list):
+        return list(
+            itertools.chain(*[_format_param(f"{key}[]", v) for v in value])
+        )
```
genevievehelsel, line 126:
```suggestion
    if isinstance(value, list):
        if not value:
            return ["-F", f"{key}[]"]
        return list(
            itertools.chain(*[_format_param(f"{key}[]", v) for v in value])
        )
```

I have not finished reading the stack, but `gh api` states "To pass an empty array, use key[] without a value." (https://cli.github.com/manual/gh_api)

if we're formatting a empty list, we should do so explicitly rather than dropping the param and not sending it since it seems like "empty array" and "missing field" can have different behaviors downstream

lazopm:
I am **really** excited for this to land! after leaving Meta having diff stacks is something I miss every day.  If we get this right it might be the thing that pushes SL more mainstream. Today it's hard for me to sell sapling to my team since the workflow generates unreadable massive PRs unless you're diligent about managing the stack, and using ReviewStack on top just feels hacky vs something native to github.

@raydatray  any chance you could upload a screencast of the workflow in your test plan? In fact, can we make **THIS** stack use the github stacked PRs feature? 

I hope we can expedite merging this behind an experimental setting flag. @genevievehelsel @raydatray

I'm writing a ton of PRs across 2-3 repos with sapling, happy to beta test and file/fix bugs as I find them.

eden/scm/sapling/ext/github/github_gh_cli.py
raydatray: hi Genevieve! thanks for taking the time to take a look, sorry for the turn around time work has been busy. i've taken another sweep at the stack and applied your suggestion

raydatray:
@lazopm 
> @raydatray any chance you could upload a screencast of the workflow in your test plan? In fact, can we make THIS stack use the github stacked PRs feature?

unfortunately since the stack feature doesnt support forks, i cant make show off the feature here. but in the subsequent diffs in the test plan i do show screenshots of me using the feature. i see that youve also done your own implementation, if you have more free time than me, i am more than happy handing this feature off to you and dogfooding it

edit: this diff shows step by step the feature https://github.com/facebook/sapling/pull/1397

Issue #1422, opened 2026-08-26
sl.exe commit -m 'added abc.txt' abort: The system cannot find the file specified: journal


swithinchan:
sl.exe commit -m 'added abc.txt' 
returns:
abort: The system cannot find the file specified: journal

OS: Windows
Steps to reproduce:
sl.exe status
A abc.txt
A ascii-home
! csv-count
! csv-filter
! csv-head
! csv-select

sl.exe commit -m 'added abc.txt'
abort: The system cannot find the file specified: journal

swithinchan:
Summary: cannot commit in Microsoft Windows at all.

WongJohnson:
I found that if you go into your .sl/store folder and create a blank file named "journal", I was able to commit. However the file is deleted afterwards and you run into the same problem for the next commit.
Reverting to this previous release worked for me: https://github.com/facebook/sapling/releases#release-0.2.20260317-201835+0234c21f

swithinchan:
Do you mean .git mode or .sl mode? I am almost abandoning this package.

WongJohnson:
sl mode. I think git mode has a .git/sl/store folder instead. Either way, try the release version I linked to see if it avoids your issue for now.

swithinchan:
Thanks @WongJohnson , as a matter of fact I have written an emacs package called sapling.el and vc-sapling.el backend in emacs for Microsoft Windows and pushed to nongnu repository. I was directed by emacs maintainers that I should follow up this issue as far as possible.

WongJohnson:
See linked PR #1431 which was landed. A build with that code should resolve the error.

swithinchan:
i cannot compile rust here. nor do i need to compile an eden server now.
@WongJohnson  is it possible to make it a release to test it out?

WongJohnson:
I'm not involved in the release process so I don't know the schedule or how to test. Were you able to try the older release I suggested previously? This one works for me: https://github.com/facebook/sapling/releases#release-0.2.20260317-201835+0234c21f
Or is that not an option for your setup?

PR #1431, opened 2026-09-13
transaction: check store journal with exists() instead of stat()

WongJohnson:
Summary:
I ran into the same issue as reported in https://github.com/facebook/sapling/issues/1422 when trying to commit in Windows, for release [0.2.20260811-150444+8fb02b32](https://github.com/facebook/sapling/releases#release-0.2.20260811-150444+8fb02b32), Aug 11. After manually creating a blank journal file, the commit was successful. (FYI release [0.2.20260317-201835+0234c21f](https://github.com/facebook/sapling/releases#release-0.2.20260317-201835+0234c21f), Mar 17 was the latest release that did not exhibit this error).

The issue was attributed to D105182535, which switched vfs to rust vfs. Previously vfs.stat mapped to os.stat, which returned the expected `FileNotFoundError`. Currently, with vfs.stat -> vfs.lstat -> bindings.io.vfs.metadata, the error returned in Windows is an `OSError` that does not get picked up by the try block.

To simplify, avoid calling svfs.stat, whose return value was used in D33482320 but is now discarded. Testing svfs.exists (which was the pre-D33482320 check, and used by `recover`) is the more direct path to triggering `recover` (or simply continuing if not needed) and should resolve the issue.

Differential Revision: D119830121


swithinchan:
i cannot compile rust here. nor do i need to eden server. 
@WongJohnson is it possible to make it a release to test it out?

